PR #93, opened 2021-05-27
setup throwaway playground regtest for e2e testing

### LINKED ISSUE #9: throw away upstream regtest capability

Instead of using TestNet/MainNet, RegTest can be used as an alternative fast network for transient based testing.

#### Why is this needed:

Enable close to TestNet testing capability without the lag time.

fuxingloh:

/kind chore

- Setup playground in CI for cypress e2e testing
- Updated `@defichain` dependencies
- Fixed workflows to use Node 15 for NPM 7
- Added `useDeFiPlayground` and `useDeFiWhale` hooks for loading client async
- Refactored `App.web.tsx` into `screens/Main.*` so that we don't repeat code
- Refactored bad hooks implementation to be cleaner
- Refactored code for better abstraction
- Added Playground.Connection with connection, provider and block count

Closes #9


fuxingloh:
@thedoublejay going to merge this in first, I want to work on other features using the playground.

Issue #57, opened 2021-05-21
`Wallet` - Encryption and Passphrase Flow


thedoublejay:
#### What would you like to be added:
Wallets are locked via passphrase. We need a way to unlock it.
#### Why is this needed:
Unlocking a wallet can be done when switching wallets or when you open the app.
Pending designs

PR #120, opened 2021-06-06
`Settings` UI Implementation 

### LINKED ISSUE #56: `settings` - main page
Settings page with some features
#### Why is this needed:
- [ ] Network Switcher
- [ ] Exit Wallet
- [ ] Language Selector

image

fuxingloh:

/kind feature

Implemented settings page.

1. Network switcher will not be enabled in early builds as it will all point to the playground for now.
2. Security will be implemented in #57 
3. All advanced options are not compatible with the light implementation
4. Language is a system-level protocol, so we are not implementing that. We will let the OS handle that.

Help & Support is wasn't present when I worked on this screen. Please open another ticket for that.
Please note that screens look different on each platform.

Fixes #56

### DIFF
--- a/screens/WalletNavigator/screens/SettingsScreen/SettingsScreen.tsx
+++ b/screens/WalletNavigator/screens/SettingsScreen/SettingsScreen.tsx
@@ -86,7 +86,10 @@ function RowNetworkItem (network?: NetworkName): JSX.Element {
 
 function RowExitWalletItem (onPress: () => void): JSX.Element {
   return (
-    <TouchableOpacity onPress={onPress} style={tailwind('bg-white')}>
+    <TouchableOpacity
+      testID='setting_exit_wallet'
+      onPress={onPress} style={tailwind('bg-white')}
+    >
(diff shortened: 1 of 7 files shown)


screens/WalletNavigator/screens/SettingsScreen/SettingsScreen.tsx
```diff
@@ -1,36 +1,98 @@
...
+        sections={[
+          {
+            key: 'Network',
+            data: [''],
+            renderItem (): JSX.Element {
+              return RowNetworkItem(network)
+            }
+          },
+          {
+            data: ['EXIT WALLET'],
```
thedoublejay, line 29: Is this label going to be translated?

thedoublejay (changes requested):

screens/WalletNavigator/screens/SettingsScreen/SettingsScreen.tsx
fuxingloh: In the end, this wasn't used. I think changed how it was designed halfway through.

thedoublejay (approved):

PR #123, opened 2021-06-07
`Playground` - E2E Tools Setup

fuxingloh:

/kind feature

- Top up testing wallet with UTXO - DFI from the playground
- List all available tokens connected to the playground instance
- Top up wallet with tokens from the playground

We don't need the design team to look through this, these are merely testing tools for e2e tests.


fuxingloh:
I am going to merge this in early first, as this utility is key for building other screens.

PR #129, opened 2021-06-08
`loader` - main loader

### LINKED ISSUE #53: `loader` - main loader
Loader component
#### Why is this needed:
Must have loading icon and injectable text (so it can be reused)
image

thedoublejay:

/kind feature

Creates a loading component that can be reused. It accepts an optional loading message (fallback to Loading). The loading layout is broken on Web view for some reason but works well on native app. 

Added temporary button on Wallet Setup page for testing in mobile.

Fixes #53


screens/LoadingScreen/LoadingScreen.tsx
```diff
@@ -0,0 +1,36 @@
+import * as React from 'react'
+import Spinner from 'react-native-loading-spinner-overlay'
+import tailwind from 'tailwind-rn'
+import { View } from '../../components/Themed'
+import { translate } from '../../translations'
+import { StyleSheet } from 'react-native'
+
+interface LoadingScreenProps {
+  message?: string
+}
```
fuxingloh, line 10: don't declare props interface if it's just one variable, more cognitive load to read and understand code in long term maintainability

fuxingloh (changes requested):
Some changes to note, you might don't want to edit the Balances screen as I am going to change it very soon. Likely conflict.

package.json
```diff
@@ -50,6 +50,7 @@
     "react-dom": "16.13.1",
     "react-native": "https://github.com/expo/react-native/archive/sdk-41.0.0.tar.gz",
     "react-native-gesture-handler": "~1.10.2",
```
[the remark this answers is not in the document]
thedoublejay: What do you mean by individually? Like a dedicated PR for this?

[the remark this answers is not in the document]
fuxingloh: Yeah, cause you updated all the deps. 

This automation will revert them daily https://github.com/DeFiCh/wallet/pull/117/files
It's not compatible with expo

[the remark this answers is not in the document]
fuxingloh: For this case, to separate out the redux PR into one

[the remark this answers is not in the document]
thedoublejay: But wait, it's not an existing dependency. It's a new package for my loader to work

screens/LoadingScreen/LoadingScreen.tsx
thedoublejay: @fuxingloh it fails the TS validation due to having infer `any` type if we remove the interface.

screens/LoadingScreen/LoadingScreen.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+const styles = StyleSheet.create({
+  loaderContainer: {
+    height: '100%',
+    width: '100%',
+    position: 'relative',
+    flex: 1,
+    justifyContent: 'center',
+    alignItems: 'center'
```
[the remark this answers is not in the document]
thedoublejay: updated

thedoublejay:
> Some changes to note, you might don't want to edit the Balances screen as I am going to change it very soon. Likely conflict.

Ok I have replaced them with your changes

package-lock.json
```diff
@@ -41,6 +41,7 @@
         "react-dom": "16.13.1",
```
fuxingloh, line 41: I meant the 1331 LOC changes

thedoublejay: ah it was re-created due to conflicts. Let me run them again

screens/LoadingScreen/LoadingScreen.tsx
fuxingloh:
```ts
export default function LoadingScreen (props: {message?: string}): JSX.Element {
```

thedoublejay: Wouldn't it look better with an interface? Especially if we want to add more params in the future?

thedoublejay: Example on why we would want to have an interface even for one variable. - https://github.com/DeFiCh/wallet/blob/818be79b6cc587df1ea6fb7b810bd78680237e2c/screens/ErrorBoundary/ErrorBoundary.tsx#L24

We declare an interface for the return type and method return types (`getDerivedStateFromError`) and in props.

fuxingloh: > Wouldn't it look better with an interface? Especially if we want to add more params in the future?

Not really, best not to do early premature optimization, especially when we have no idea of how it will evolve.

fuxingloh: > Example on why we would want to have an interface even for one variable. -
> 
> https://github.com/DeFiCh/wallet/blob/818be79b6cc587df1ea6fb7b810bd78680237e2c/screens/ErrorBoundary/ErrorBoundary.tsx#L24
> 
> We declare an interface for the return type and method return types (`getDerivedStateFromError`) and in props.

I think more and more conventions are written in use hooks now, might not be relevant.

thedoublejay: Ok I will update. 

For `getDerivedStateFromError` there's no hooks for it yet. See https://reactjs.org/docs/hooks-faq.html#do-hooks-cover-all-use-cases-for-classes

fuxingloh (commented):
looks good but let's keep it drafted, for now, I can't see how RN and the full screen will behave for this.

fuxingloh (approved):

PR #136, opened 2021-06-11
Translations fix to prevent "guessing" results

thedoublejay:

/kind fix

Currently, `i18njs` behavior is to "guess" the missing translations. It is causing some issues with texts that contains punctuation marks. Also, it replaces camel case text with spaces which is an issue with some proper nouns (ex. WeChat -> We Chat). Reference - https://github.com/fnando/i18n-js#translation-missing-behaviour-control

Current behavior is to use the "text" as fallback if translation is missing.

Fixes #

### DIFF
--- a/translations/index.ts
+++ b/translations/index.ts
@@ -53,5 +53,9 @@ export function translate (path: string, text: string): string {
     initI18n()
   }
   const translation = i18n.translate(`${path}.${text}`)
-  return translation != null && translation !== '' ? translation : text
+  if (translation !== null && translation !== undefined && translation !== '') {
+    return translation
+  } else {
+    return text
+  }
 }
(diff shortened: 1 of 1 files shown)


translations/index.ts
```diff
@@ -52,5 +52,6 @@ export function translate (path: string, text: string): string {
   if (!init) {
     initI18n()
   }
-  return i18n.translate(`${path}.${text}`)
+  const translation = i18n.translate(`${path}.${text}`)
+  return translation != null && translation !== '' ? translation : text
```
fuxingloh, line 56: I prefer if condition than ternary if your has more than 1 condition, in this case your ternary has 2 conditions.

`if ()` allows for better code readability as you can clearly see where the code condition exits or continue.

fuxingloh (changes requested):

translations/index.ts
thedoublejay: updated

fuxingloh (approved):

PR #151, opened 2021-06-14
Playground access within Settings

fuxingloh:

/kind feature

Debug tools for preview testing and preview QOL.

1. You can wallet setup by holding down the wallet icon for 5 seconds.
    

2. You can access the playground within settings by clicking on the playground.
    

Also made text in Playground larger, so that it's easier to see for testing.


thedoublejay (approved):
good idea

PR #162, opened 2021-06-16
`balances` - main page

### LINKED ISSUE #50: `balances` - main page
Balances Page
#### Why is this needed:
Ability to display balances, send and receive tokens
- [ ] Balances Page
- [ ] Token List

image

thedoublejay:

/kind feature

1. Updated SVG icons to be consistent
2. Added `react-number-format` and created `NumberText` component for formatting of number 
3. Added `useTokensAPI` hook that fetches the balances and tokens from whale client. It returns the tokens from global state.
4. Fixed memory leak on `Playground.Connection.tsx` on mobile
5. Displayed list of tokens in Balances page with ability to refresh via swipe down
6. Stored `address` in global wallet state to prevent calling API on every wallet address reference

Fixes #50


store/wallet.ts
```diff
@@ -22,6 +38,29 @@ export const wallet = createSlice({
...
+    setTokens: (state, action: PayloadAction<AddressToken[]>) => {
+      state.tokens = action.payload
+    },
+    setUtxoBalance: (state, action: PayloadAction<string>) => {
+      state.utxoBalance = action.payload
     }
   }
 })
+
+export const { setStatus, setTokens, setUtxoBalance } = wallet.actions
```
fuxingloh, line 51: The reason why I didn't want to export this individually was it creates ambiguity, 

for example `setStatus` for what which action? To find out when reading and maintaining code you have to reference the imports to know what that code does. It's just one of those patterns that lead to code debt and introduces unnessary bugs.

While the long-form, `wallet.actions.setStatus` is clear and concise in what you are trying to achieve.

store/wallet.ts
```diff
@@ -22,6 +38,29 @@ export const wallet = createSlice({
...
+
+const rawTokensSelector = createSelector((state: WalletState) => state.tokens, (tokens) => {
+  if (tokens.some((t) => t.id === '0')) {
+    return tokens
+  } else {
+    return [initialDFI, ...tokens]
+  }
+})
+//* Use as token selector to include UTXO balance
+export const tokensSelector = createSelector([rawTokensSelector, (state: WalletState) => state.utxoBalance], (tokens, utxoBalance) => tokens.map((t) => {
```
fuxingloh, line 61: I think we should not merge it, display it separately. My thoughts are people constantly explaining to each other what is UTXO and Account token. We could do that within our design instead, users should not need to reference another help document (telegram) to use the app.

hooks/wallet/TokensAPI.ts
```diff
@@ -0,0 +1,46 @@
...
+import { useWhaleApiClient } from '../api/useWhaleApiClient'
+import { useWalletAPI } from './WalletAPI'
+
+const fetchWalletTokens = async (address: string, whaleAPI: WhaleApiClient): Promise<ApiPagedResponse<AddressToken>> => {
+  return await whaleAPI.address.listToken(address)
+}
+
+const fetchWalletBalance = async (address: string, whaleAPI: WhaleApiClient): Promise<string> => {
+  return await whaleAPI.address.getBalance(address)
+}
```
fuxingloh, line 17: we should not split this, should just call it directly

hooks/wallet/TokensAPI.ts
```diff
@@ -0,0 +1,46 @@
...
+  useEffect(() => {
+    let intervalId: NodeJS.Timeout
+    account.getAddress().then(async address => {
+      await fetchWalletData(address, dispatch, whaleAPI)
+      intervalId = setInterval(() => {
+        fetchWalletData(address, dispatch, whaleAPI)
+      }, 10000)
+    }).catch((error) => console.log(error))
+    return () => clearInterval(intervalId)
+  }, [])
```
fuxingloh, line 44: We have to drop the interval refreshing, at least for now, how tabs are implemented native means this will keep refreshing even if we have navigated into a deep controller.

This will reduce the complexity greatly for our code.

We will also need:
1. ability to trigger a manual refresh
2. pagination of all tokens

> Kinda awkward that this code section is here, would make more sense in redux slice but I think it is limited by `useWhaleApiClient` being a hook. In the future, we will have to refactor that for code simplicity later.

fuxingloh (commented):
looks great overall, I left some code debt and native concerns that I think you should address

store/wallet.ts
thedoublejay: Ok, only downside is we're importing the entire object.

store/wallet.ts
thedoublejay: This one will be used as a display as the main list of tokens. We can provide a split in display but somewhere we still need to have a total. Also when doing operations, it would still be better to have a total. 

But yeah we can have a breakdown, of current UTXO and Token.

hooks/wallet/TokensAPI.ts
thedoublejay: For the interval part, I think we should keep it. The reason is users tend to wait on the page when a transaction is pending. Having to tell them to manual refresh or switch tabs might not be a good solution. It might be reported as a bug/issue in the future.

store/wallet.ts
fuxingloh: don't worry about this, it's automatically optimized when it's compiled re-`export/imports` is more or less for QOL purposes

store/wallet.ts
fuxingloh: We don't need total for any operation as none of the operations allows it. I think the total is more for communicating our design. Still would be useful

hooks/wallet/TokensAPI.ts
fuxingloh: I think you misunderstand what I mean, not saying we don't need it in the future but saying we should remove it now to reduce complexity and code debt.

For reasons that we need it updated, we can trigger the updates manually.

thedoublejay:
@fuxingloh I think the PR is already huge and can be reviewed and merged first before proceeding on the next iteration. I have updated the description based on the changed files.

Pending items
- Token Pagination
- E2E tests

hooks/wallet/TokensAPI.ts
thedoublejay: I have remove the interval for tokens

hooks/wallet/WalletAPI.ts
```diff
@@ -78,6 +78,8 @@ export function useWalletAPI (): typeof WalletAPI {
     async function loadWallet (): Promise<void> {
       if (await hasMnemonicWallet()) {
         jellyfishWallet = await getMnemonicWallet(client, network as NetworkName)
+        const address = await jellyfishWallet.get(0).getAddress()
+        dispatch(wallet.actions.setAddress(address))
```
fuxingloh, line 82: not a good idea, we should not cache addresses as the wallet is agnostically provided. we will keep this for now

store/wallet.ts
```diff
@@ -22,6 +40,30 @@ export const wallet = createSlice({
...
+  if (t.id === '0') {
+    return { ...t, amount: new BigNumber(t.amount).plus(utxoBalance).toFixed() }
+  }
+  return t
+}))
```
fuxingloh, line 69: Hesitant in putting UI logic in the global state as it is only unique to the balances page.

screens/AppNavigator/screens/BalancesScreen/BalancesScreen.tsx
```diff
@@ -1,52 +1,67 @@
...
+    </View>
+  )
+}
 
+function BalanceItemRow ({ token }: { token: AddressToken }): JSX.Element {
+  const Icon = getTokenIcon(token.symbol)
   return (
-    <View style={tailwind('flex-1 items-center justify-center')}>
-      <View style={tailwind('flex-row items-center')}>
+    <View testID={token.id} style={tailwind('bg-white p-2 border-b border-gray-200 flex-row items-center w-full h-16')}>
```
fuxingloh, line 54: try to use as little style as possible as not everything is required for rendering what you want

screens/AppNavigator/screens/BalancesScreen/BalancesScreen.tsx
```diff
@@ -1,52 +1,67 @@
...
-    return (
-      <View key={token.id} style={tailwind('flex-row items-center')}>
-        <Icon />
-        <Text style={tailwind('ml-2')}>{token.amount}</Text>
-      </View>
-    )
-  })
+  const tokens = useTokensAPI()
+  return (
+    <View style={tailwind('items-start justify-start w-full flex-grow')}>
```
fuxingloh, line 29: you can remove this top View, should not be required

screens/AppNavigator/screens/BalancesScreen/BalancesScreen.tsx
```diff
@@ -1,52 +1,67 @@
...
+      <FlatList
+        style={tailwind('w-full mb-4')}
```
fuxingloh, line 31: don't use margin

fuxingloh (approved):

store/wallet.ts
thedoublejay: @fuxingloh it's not in the global state. It's only in the selector. Also, it will also be used in DEX (whenever you select a token).

PR #165, opened 2021-06-17
updated all icons to use MaterialIcons set

### LINKED ISSUE #164: Update icon set used to MaterialIcons and/or MaterialCommunityIcons
Update icon set used to MaterialIcons and/or MaterialCommunityIcons.

#### Why is this needed:
The desktop app uses Material Design Icons. The designs also specify Material Design icons. We should unify all icons used in every product for consistency. MaterialCommunityIcons are additional and non-official icons that are consistent with the MD style, and may be useful if we need something not provided by the base set.

Tab bar and anywhere else in the UI that shows icons.

All design icon specifications are in Marvel handoffs.

fuxingloh:

/kind fix

Updated existing icon set to use MaterialIcons set.

Fixes #164 

#### Additional comments?:

- I am not a big fan of using Material on non-android devices, we should use a platform-agnostic icon set. We should change the icon set in the future.
- Did not use the MaterialCommunity set as it's not a superset of Material and the icons' name and style are not consistent with the original material guidelines.

### DIFF
--- a/constants/Theme.ts
+++ b/constants/Theme.ts
@@ -1,5 +1,7 @@
+import { MaterialIcons } from '@expo/vector-icons'
 import { DefaultTheme } from '@react-navigation/native'
 import { Theme } from '@react-navigation/native/lib/typescript/src/types'
+import * as React from 'react'
 import { StyleSheet } from 'react-native'
 
 export const PrimaryColor = '#ff00af'
@@ -20,3 +22,6 @@ export const DeFiChainTheme: Theme = {
     primary: PrimaryColor
   }
 }
+
+export const VectorIcon = MaterialIcons
+export type VectorIconName = React.ComponentProps<typeof MaterialIcons>['name']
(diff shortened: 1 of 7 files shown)


aentan:
There's no such thing as a strictly platform-specific icon set. Even the native icon set by iOS isn't that comprehensive. Almost all non-Apple apps uses their own icon set. I chose Material because it's comprehensive, easy to design new and more custom icons around a simple grid. Plus consistency with the desktop app already using them.

In the end, most end users wouldn't consciously care, just subconsciously pick up hints of UI dysfunction. At this point I've evaluated that MD is the most comprehensive set for use across iOS, Android and web.

fuxingloh:
> There's no such thing as a strictly platform-specific icon set. Even the native icon set by iOS isn't that comprehensive. Almost all non-Apple apps uses their own icon set. I chose Material because it's comprehensive, easy to design new and more custom icons around a simple grid. Plus consistency with the desktop app already using them.

Actually, that's my point, I am going with platform-agnostic not platform-specific.

thedoublejay (approved):

PR #179, opened 2021-06-21
`balances` page tests

thedoublejay:

/kind chore

Unit tests and e2e test for balances page.
1. Added `fetch balances` button for Playground to get latest tokens
2. Still haven't added test for "Pull to refresh". Having trouble simulating the swipe down event

Fixes #


cypress/integration/functional/balances.spec.ts
```diff
@@ -0,0 +1,51 @@
...
+    tokens.forEach((item) => {
+      const baseTestID = `balances_row_${item.id}`
+      flatList.getByTestID('balances_title').should('exist').contains('Portfolio')
+      flatList.getByTestID(baseTestID).should('exist')
+      flatList.getByTestID(`${baseTestID}_icon`).should('exist')
+      flatList.getByTestID(`${baseTestID}_symbol`).should('exist').contains(item.symbol)
+      flatList.getByTestID(`${baseTestID}_name`).should('exist').contains(item.name)
+      flatList.getByTestID(`${baseTestID}_amount`).should('exist').contains(item.amount)
+    })
+  })
```
fuxingloh, line 50: instead of doing this, can write the test more loosely? this way of testing makes it hard to write evolve components.

cypress/integration/functional/balances.spec.ts
```diff
@@ -0,0 +1,51 @@
...
+  {
+    id: '1',
+    symbol: 'tBTC',
+    amount: '100',
+    name: 'Playground BTC',
+    isDAT: true,
+    isLPS: false,
+    symbolKey: 'tBTC'
+  }
+]
```
fuxingloh, line 22: don't write a test like this, stub based testing is harder to maintain in the long run use a more functional approach where you actually write out the test contract instead of looping them through.

A test should be more explicit with fewer references and inferences. It should be clearly explicit.

fuxingloh (changes requested):

cypress/integration/functional/balances.spec.ts
thedoublejay: ok understood, will update

cypress/integration/functional/balances.spec.ts
thedoublejay: I have removed this

cypress/integration/functional/balances.spec.ts
thedoublejay: I have simplified it to check if the rows exists and the amount is correct

cypress/integration/functional/balances.spec.ts
```diff
@@ -0,0 +1,29 @@
+context('wallet/balances', () => {
+  beforeEach(function () {
+    cy.createEmptyWallet(true)
+    const baseURL = '/v1/playground/rpc'
+    cy.intercept(`${baseURL}/sendtoaddress`).as('sendToAddress')
+    cy.intercept(`${baseURL}/sendtokenstoaddress`).as('sendTokensToAddress')
+    cy.getByTestID('bottom_tab_settings').click()
+    cy.getByTestID('playground_wallet_top_up').click()
+    cy.getByTestID('playground_token_tBTC').click()
+    cy.wait(['@sendToAddress', '@sendTokensToAddress'])
```
fuxingloh, line 10: I think it would be awesome if you could write a command first this in crypress

cypress/integration/functional/balances.spec.ts
```diff
@@ -0,0 +1,29 @@
...
+    // Check if title exists
+    cy.getByTestID('balances_title').should('exist').contains('Portfolio')
+    // Check if DFI exists
+    cy.getByTestID('balances_row_0').should('exist')
+    // Check if tBTC exists
+    cy.getByTestID('balances_row_1').should('exist')
+    // Check if DFI amount is correct
+    cy.getByTestID('balances_row_0_amount').contains(50)
+    // Check if tBTC amount is correct
+    cy.getByTestID('balances_row_1_amount').contains(100)
```
fuxingloh, line 27: actually, you don't need to write the comment above since the test code is quite obvious what you are testing

fuxingloh (approved):
thanks for the change! I will update the UI to match the design more closely later. I needed these changes because your tests were too strict for evolving UI codes

cypress/integration/functional/balances.spec.ts
thedoublejay: ok I will update it on a new PR

fuxingloh:
I am going to merge this first

PR #184, opened 2021-06-21
DEX `Your PoolPair`

fuxingloh:

/kind feature

Fixes #


thedoublejay:
@fuxingloh I just realized. This page should be Liquidity instead of DEX right?

fuxingloh:
> @fuxingloh I just realized. This page should be Liquidity instead of DEX right?

Yeah, I reshuffled the names. DEX makes more sense since liquidity is quite ambiguous. I am going to move all DEX features (swap/add/remove) into that screen since they belong together. Might rename it to Decentralized Exchange too, we shall see!

thedoublejay (approved):

PR #191, opened 2021-06-23
`transaction`: listing page e2e tests

ivan-zynesis:

/kind feature

e2e test cases

Fixes #


cypress/integration/functional/transaction/list.spec.ts
```diff
@@ -0,0 +1,29 @@
...
+      // go to page for testing
+      cy.getByTestID('bottom_tab_transactions').click()
+    })
+
+    it('should display 2 rows', () => {
+      cy.getByTestID('transactions_screen_list').should('exist')
+      cy.getByTestID('transaction_row_0').should('exist')
+      cy.getByTestID('transaction_row_1').should('exist')
+      // seems to be crashing (when trying to get element if testid not found)
+      // cy.getByTestID('transaction_row_2').should('not.exist')
```
ivan-zynesis, line 26: @thedoublejay this is what I mean by finding "expect to be not exist" element test fail, and it will cause the test time to be extreme long before it fail. I will disable it first until we can investigate further.

thedoublejay: Ok will check also. Syntax wise looks ok, maybe the usage of custom command breaks it but to be checked

thedoublejay (approved):

fuxingloh (approved):

PR #242, opened 2021-07-08
added colors for IconDefault

thedoublejay:

/kind feature

Added colors for IconDefault. This will help us differentiate tokens. Tokens are consistent across all pages and unique based on their symbol.

Fixes #

#### Additional comments?:
<div>

</div>


ivan-zynesis (approved):
Interest constant to color snippet

app/components/icons/tokens/IconDefault.tsx
```diff
@@ -1,13 +1,20 @@
...
+import { useEffect, useState } from 'react'
 import Svg, { Circle, SvgProps, Text } from 'react-native-svg'
 
 export function IconDefault (symbol: string): (props: SvgProps) => JSX.Element {
   return (props: SvgProps): JSX.Element => {
+    const [color, setColor] = useState('')
+    useEffect(() => {
+      const c = randomColor({ luminosity: 'dark', seed: symbol })
+      setColor(c)
+    }, [symbol])
```
fuxingloh, line 12: instead of random, it should be hash-based color derived from the token name and being deterministic

---

And assuming the above is true since your random is deterministic you don't need to `useEffect` and you can just use do a normal function call?

fuxingloh (changes requested):

app/components/icons/tokens/IconDefault.tsx
thedoublejay: Ah yeah good point, I'll remove the `useEffect`. it's already based on the token name (seed option)

app/components/icons/tokens/IconDefault.tsx
```diff
@@ -1,13 +1,15 @@
...
+    const color = randomColor({ luminosity: 'dark', seed: symbol })
     return (
       <Svg width={32} height={32} viewBox='0 0 32 32' {...props}>
-        <Circle cx={16} cy={16} r={16} fill='#444444' />
+        <Circle cx={16} cy={16} r={16} fill={color} />
         <Text
-          x='50%' y='50%' alignmentBaseline='middle' textAnchor='middle' fontWeight='bolder' fontSize='11'
+          x='50%' y='50%' alignmentBaseline='middle' textAnchor='middle' fontWeight='bolder' fontSize='8'
           fill='#ffffff'
         >{symbol}
```
fuxingloh, line 14: I think we can just use the first letter, following the designed materials from Aen

thedoublejay: updated

thedoublejay:
> Interest constant to color snippet

What do you mean by this @ivan-zynesis?

fuxingloh:
> > Interest constant to color snippet
> 
> What do you mean by this @ivan-zynesis?

I am guessing "increase"?

fuxingloh (approved):

PR #252, opened 2021-07-09
`OceanInterface` to display ongoing network calls/transactions and `send` api

### LINKED ISSUE #201: `Bottom Drawer` for displaying transaction and network activity

Instead of using the navbar which is size fragmented and has a different design language in each native platform. We should use Bottom Tab Bar and have a Drawer/Fold above that tab bar instead.

/area ui-ux

### LINKED ISSUE #249: Tab bars need labels

Tab bars should have labels.

#### Why is this needed:

It's hard to know what each does is for without labels.

Tab bar.

Design has labels. See design for handoff specs. e.g. https://marvelapp.com/prototype/c572fa6/screen/78750417/handoff

thedoublejay:

/kind feature

1. Global component that can be use to display ongoing network calls/transaction ids. Please refer to the video below.
2. Network Drawer has 3 states. Error, Loading, Success
3. Added Send Logic for Token and DFI
4. Ability to open explorer
5. Display fee in Send Screen
6. Added smart-buffer for unit test
7. On failed broadcast attempt, there's 1 retry for it

Fixes #201

#### Additional comments?:

https://user-images.githubusercontent.com/25013382/125435359-b3d00959-462a-4c5e-bce1-1440059c4b45.mp4


app/store/networkDrawer.ts
```diff
@@ -0,0 +1,30 @@
+import { createSlice, PayloadAction } from '@reduxjs/toolkit'
+
+export interface NetworkDrawerState {
+  isLoading: boolean
+  isOpen: boolean
+  height: number
+  title: string
+  txid?: string
```
fuxingloh, line 8: I don't think you should not track component state in store (UI state), you should only track data

In this case, pending transaction data

fuxingloh (changes requested):

app/store/networkDrawer.ts
thedoublejay: Instead of using store for UI state, we need to pass status of transaction? then we just derive the UI state from that. Am I correct?

fuxingloh: Ideally, the scenario would be:

1. push signed transaction to the drawer (global state)
2. drawer will enqueue the transaction and send the whale API (if online)
3. drawer will spin loop until and transaction gets confirmed therefore removing it from the drawer

Something like this: 
```ts
interface State {
  transactions: {
    broadcasted: boolean
    signed: CTransactionSegWit[]
  }[]
}
```

thedoublejay: With this approach, I feel like we're making it coupled with transactions. Wouldn't it be better to make it a dumb component and let the called component dictate when/what to display this component? So it will be flexible enough to use on other use cases as well. (e.g, fetching data, displaying errors etc.)

ivan-zynesis: If that's the case, the transactions.signed will behaves like a queue. The drawer will handle the all the submission, status, error rendering (within drawer component)?

fuxingloh: Yeah it should behave like a queue for the blockchain. The entire idea of that component is to add as a middleware for connecting to the blockchain.

Be it network status or bridging transactions.

app/screens/AppNavigator/BottomTabNavigator.tsx
```diff
@@ -27,44 +30,50 @@ function TabBarIcon (props: { name: VectorIconName, color: string }): JSX.Elemen
 }
 
 export function BottomTabNavigator (): JSX.Element {
+  const isNetworkDrawerOpen = useSelector((state: RootState) => state.networkDrawer.isOpen)
   return (
-    <BottomTab.Navigator
-      initialRouteName='Balances'
-      tabBarOptions={{ adaptive: false, showLabel: false }}
```
fuxingloh, line 33: I am going to open another PR for `showLabel: fix` to get it in first and also fix some UI consistency that was reported

thedoublejay: Ok copy

ivan-zynesis (changes requested):

app/store/networkDrawer.ts
```diff
@@ -0,0 +1,51 @@
+import { CTransactionSegWit } from '@defichain/jellyfish-transaction'
+import { createSelector, createSlice, PayloadAction } from '@reduxjs/toolkit'
+
+export interface Transaction {
```
ivan-zynesis, line 4:
```suggestion
export interface TransactionJob {
```

For not to confused with jellyfish existing Transaction interface

app/store/networkDrawer.ts
```diff
@@ -0,0 +1,51 @@
...
+  broadcasted: boolean
+  signed: CTransactionSegWit
+  title?: string
```
ivan-zynesis, line 7:
```suggestion
  title?: string
  err?: Error
```

I think will be better to keep for each job, possible to retry too.

app/components/NetworkDrawer/NetworkDrawer.tsx
```diff
@@ -0,0 +1,146 @@
...
+          ...transaction,
+          title: translate('screens/NetworkDrawer', 'Transaction complete'),
+          broadcasted: true
+        })
+        dispatch(networkDrawer.actions.popTransaction())
+      }).catch(() => {
+      })
+    }
+  }, [transaction])
+
```
ivan-zynesis, line 65:
```suggestion
  if (transaction === undefined) return null
```
By doing this we can remove all the `setHeight()` right?

app/store/networkDrawer.ts
```diff
@@ -0,0 +1,51 @@
...
+  name: 'networkDrawer',
+  initialState,
+  reducers: {
+    setHeight: (state, action: PayloadAction<number>) => {
+      state.height = action.payload
+    },
+    queueTransaction: (state, action: PayloadAction<Transaction>) => {
+      state.transactions = [...state.transactions, action.payload]
+    },
+    closeNetworkDrawer: (state) => {
```
ivan-zynesis, line 32: Where is this applied? This will force cancel ongoing task, can avoid it if not necessary. I think even `popTransaction` should be consumed by NetworkDrawer only.

app/screens/AppNavigator/BottomTabNavigator.tsx
```diff
@@ -27,44 +31,50 @@ function TabBarIcon (props: { name: VectorIconName, color: string }): JSX.Elemen
...
-        component={SettingsNavigator}
-        options={{
-          tabBarTestID: 'bottom_tab_settings',
-          tabBarIcon: ({ color }) => <TabBarIcon name='settings' color={color} />
-        }}
-      />
-    </BottomTab.Navigator>
+    <>
+      {
+        isNetworkDrawerOpen && <NetworkDrawer />
```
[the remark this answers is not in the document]
thedoublejay: ok thanks, will test them first on mobile

app/components/NetworkDrawer/NetworkDrawer.tsx
```diff
@@ -0,0 +1,146 @@
...
+async function broadcastTransaction (tx: Transaction, client: WhaleApiClient, setError: (e: Error) => void): Promise<string | undefined> {
+  let retries = 0
+  try {
+    return await client.transactions.send({ hex: tx.signed.toHex() })
+  } catch (e) {
+    retries++
+    Logging.error(e)
+    if (retries < 1) {
+      return await broadcastTransaction(tx, client, setError)
```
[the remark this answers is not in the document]
thedoublejay: it doesn't. Retry value will be same due to closure. I have tested it on my error

app/components/NetworkDrawer/NetworkDrawer.tsx
thedoublejay: Nope, setHeight is required to have exact measurements of bottom drawer. It's unrelated to transaction but it's merely a UI state. Height varies per device so we need to set it to achieve the style of drawer

app/store/networkDrawer.ts
thedoublejay: It's applied on the "OK" button if it's the last transaction of the stack.

app/contexts/NetworkContext.tsx
```diff
@@ -11,6 +12,18 @@ interface Network {
 
 const NetworkContext = createContext<Network>(undefined as any)
 
+export function networkMapper (network: EnvironmentNetwork): NetworkName {
```
fuxingloh, line 15: don't edit the context to export network config out, you are pulling network config out instead of reading into it.

Reverse the flow of this implementation, such that you expose this instead: `const { parseAddress } = useNetworkContext()`

app/contexts/WhaleContext.tsx
```diff
@@ -1,7 +1,7 @@
 import { WhaleApiClient } from '@defichain/whale-api-client'
 import React, { createContext, useContext, useMemo } from 'react'
 import { EnvironmentNetwork } from '../environment'
-import { useNetworkContext } from './NetworkContext'
+import { networkMapper, useNetworkContext } from './NetworkContext'
```
fuxingloh, line 4: Also to revert this.

fuxingloh (changes requested):
Should call this `OceanInterface` since this component interface with the ocean ecosystem.

app/store/networkDrawer.ts
thedoublejay: I'll rename it OceanTransaction since it will be called OceanInterface now

app/screens/AppNavigator/BottomTabNavigator.tsx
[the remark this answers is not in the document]
thedoublejay: I have tried this and unfortunately it doesn't work. It throws some rendering error on web and mobile

app/store/networkDrawer.ts
thedoublejay: I'll keep this for now. I don't see the reason of storing the error per transaction since we only display one at a time. Currently, there's one error on OceanInterface.

app/store/index.ts
```diff
@@ -14,8 +15,12 @@ import { wallet } from './wallet'
 export const store = configureStore({
   reducer: {
     block: block.reducer,
-    wallet: wallet.reducer
-  }
+    wallet: wallet.reducer,
+    oceanInterface: oceanInterface.reducer
```
fuxingloh, line 19:
```suggestion
    ocean: ocean.reducer
```

I think we can just call this `ocean`

app/contexts/NetworkContext.tsx
fuxingloh:
```suggestion
function networkMapper (network: EnvironmentNetwork): NetworkName {
```

don't export this

fuxingloh (changes requested):

app/screens/AppNavigator/screens/Balances/BalancesScreen.tsx
```diff
@@ -11,18 +12,24 @@ import { getTokenIcon } from '../../../../components/icons/tokens/_index'
...
 export function BalancesScreen ({ navigation }: Props): JSX.Element {
+  const height = useBottomTabBarHeight()
   const client = useWhaleApiClient()
   const address = useSelector((state: RootState) => state.wallet.address)
   const [refreshing, setRefreshing] = useState(false)
   const dispatch = useDispatch()
 
+  useEffect(() => {
+    dispatch(ocean.actions.setHeight(height))
+  }, [height])
```
fuxingloh, line 31: slightly uncomfortable with this height injection, hope we can refactor in the future

fuxingloh (approved):

app/screens/AppNavigator/screens/Balances/BalancesScreen.tsx
thedoublejay: yeah, I'll check if we can have static numbers for it or if we have a better way to get exact height. I'll create a ticket for it

PR #270, opened 2021-07-14
Combined Unit + E2E Coverage

### LINKED ISSUE #247: Collect CodeCoverage from E2E Cypress Tests

Non UI code was tested with jest unit test, other tests such as those that require navigation uses Cypress for testing. Coverage can be called from Cypress suite and we should collect them to represent the collective coverage we did for `DeFiCh/wallet`.

thedoublejay:

/kind chore

Fixes #247 

#### Additional comments?:

- Added some Cypress config for better testing QOL
- Codecov will merge coverage reports automatically so we can just use that
- Cleanup docs and health file-related configs


fuxingloh:
- Added some Cypress config for better testing QOL
- Codecov will merge coverage reports automatically so we can just use that
- Cleanup docs and health file-related configs

fuxingloh (approved):

PR #280, opened 2021-07-15
ocean interface store action improvement

ivan-zynesis:

/kind feature

- Fix broadcast retry logic
- Trim down unnecessary OceanInterface global state, ease setup on new page
- Improve lifecycle within OceanInterface component
- Tighten up the interface to interact with OceanInterface
- Also added all job queue checker before create new DfTx (probably should open separate PR, if find any blocker for this to merge)
- Move main signing logic execution to be ran inside OceanInterface
- Drawer will display error message instead if signing fail (no txid found)

Fixes #


app/components/OceanInterface/OceanInterface.tsx
```diff
@@ -21,61 +23,78 @@ async function handlePress (txid: string): Promise<void> {
...
+        broadcasted: false
       })
+
+      broadcastTransaction(transaction, client)
+        .catch((e: Error) => setError(e))
+        .finally(() => dispatch(ocean.actions.popTransaction())) // remove the job as soon as completion
     }
   }, [transaction])
 
+  if (tx === undefined) {
```
ivan-zynesis, line 77: `tx` is the ONE and ONLY must have state to render the component. Logic/hooks before this line should control it properly.

app/components/OceanInterface/OceanInterface.tsx
```diff
@@ -21,61 +23,78 @@ async function handlePress (txid: string): Promise<void> {
...
+        err !== undefined
+          ? <TransactionError txid={tx.signed.txId} onClose={dismissDrawer} />
+          : <TransactionDetail tx={tx} onClose={dismissDrawer} />
+      }
     </Animated.View>
   )
 }
 
-function TransactionDetail ({ tx }: { tx: OceanTransaction }): JSX.Element {
+function TransactionDetail ({ tx, onClose }: { tx: OceanTransaction, onClose: () => void }): JSX.Element {
```
ivan-zynesis, line 97: remove `dispatch` pass on to all sub-components in this page, isolate the logic and renderer

app/screens/AppNavigator/BottomTabNavigator.tsx
```diff
@@ -31,12 +28,9 @@ function TabBarIcon (props: { name: VectorIconName, color: string }): JSX.Elemen
 }
 
 export function BottomTabNavigator (): JSX.Element {
-  const isOceanInterfaceOpen = useSelector((state: RootState) => isDrawerOpenSelector(state.ocean))
   return (
     <>
-      {
-        isOceanInterfaceOpen && <OceanInterface />
-      }
+      <OceanInterface />
```
ivan-zynesis, line 33: The issue was component attempt to render when the queue is empty. Should be okay now.

app/components/OceanInterface/OceanInterface.tsx
```diff
@@ -11,71 +11,98 @@ import { useWhaleApiClient } from '../../contexts/WhaleContext'
...
   } catch (e) {
-    retries++
     Logging.error(e)
-    if (retries < 1) {
-      return await broadcastTransaction(tx, client, setError)
+    if (retries < MAX_AUTO_RETRY) {
+      return await broadcastTransaction(tx, client, retries + 1)
     }
-    setError(e)
+    throw e
```
ivan-zynesis, line 41: Error thrown after 2 attempts

app/components/OceanInterface/OceanInterface.tsx
```diff
@@ -11,71 +11,98 @@ import { useWhaleApiClient } from '../../contexts/WhaleContext'
...
-        dispatch(ocean.actions.popTransaction())
-      }).catch(() => {
+      console.log('broadcasting', transaction.signed.txId)
+      setTx({
+        ...transaction,
+        broadcasted: false
       })
+
+      broadcastTransaction(transaction, client)
+        .then(() => setTx({ ...transaction, broadcasted: true, title: translate('screens/OceanInterface', 'Sent') }))
```
ivan-zynesis, line 79: Update the local UI state to display success tx broadcast. UI will be replaced by next transaction if any in the queue.

app/store/ocean.ts
```diff
@@ -29,25 +29,16 @@ export const ocean = createSlice({
...
-    },
     setError: (state, action: PayloadAction<Error>) => {
       state.err = action.payload
     },
     popTransaction: (state) => {
-      if (state.transactions.length > 1) {
-        state.transactions.shift()
-        state.transactions = [...state.transactions]
-      }
+      state.transactions.shift()
```
ivan-zynesis, line 36: This solely serves as job queue. Remove them all even is last one. UI kept the last in state to display until manual dismissed.

app/store/ocean.ts
```diff
@@ -3,7 +3,7 @@ import { createSelector, createSlice, PayloadAction } from '@reduxjs/toolkit'
 
 export interface OceanTransaction {
   broadcasted: boolean
-  signed: CTransactionSegWit
+  signer: () => Promise<CTransactionSegWit>
```
ivan-zynesis, line 6: Major change here @thedoublejay.
I moved the signing logic to be executed inside OceanInterface component. The signing itself is taking way longer than the broadcast, the UI looks frozen before the dispatch(queue) happened and kinda defeat the purpose of having this drawer.

thedoublejay: yup good idea

app/components/OceanInterface/OceanInterface.tsx
```diff
@@ -11,94 +11,129 @@ import { useWhaleApiClient } from '../../contexts/WhaleContext'
...
-async function handlePress (txid: string): Promise<void> {
-  // TODO(thedoublejay) explorer URL
-  const url = `https://explorer.defichain.io/#/DFI/mainnet/tx/${txid}`
-  const supported = await Linking.canOpenURL(url)
-  if (supported) {
-    await Linking.openURL(url)
-  }
+const MAX_AUTO_RETRY = 1
+
+// async function gotoExplorer (txid: string): Promise<void> {
```
thedoublejay, line 19: I think we can keep the redirect and store the txid in the clipboard momentarily.

app/components/OceanInterface/OceanInterface.tsx
thedoublejay: thanks for this, my bad for this bug

app/components/OceanInterface/OceanInterface.tsx
thedoublejay: Let's keep the copy as "Transaction complete"/"Transaction sent". "Sent" can mean many things and quite be confusing for different operations (DEX, LM etc.)

app/components/OceanInterface/OceanInterface.tsx
```diff
@@ -107,36 +142,33 @@ function TransactionError ({ txid }: { txid: string | undefined }): JSX.Element
           style={tailwind('text-sm font-bold')}
         >{`${translate('screens/OceanInterface', 'An error has occurred')}`}
         </Text>
-        {
```
thedoublejay, line 110: I think we shouldn't remove this TxID. The idea of this Transaction error is to display if there's any JS errors or if there's an error related to the transaction.

Unless there's no case where a txid is broadcasted and an error will occur. We can remove this

app/screens/AppNavigator/BottomTabNavigator.tsx
thedoublejay: This might be the cause also why the animation doesn't work anymore. I'll check on how we can move it on the new structure.

thedoublejay (changes requested):

app/components/OceanInterface/OceanInterface.tsx
ivan-zynesis: As now signing itself as a promise can fail, so txid is not always available. But make sense to display it if we have one. Will include it in the message crafting if signing did went through.

thedoublejay (approved):

PR #295, opened 2021-07-16
ui fixes

### LINKED ISSUE #290: `balances` - device with larger fonts breaks layout
#### What happened:
On devices which are using larger fonts, the layout breaks.
Should wrap text properly

Larger font on Settings vs Normal font
<div>

</div>

thedoublejay:

/kind fix

1. Fixed text overflow on Balances Page
2. Fixed animation and text overflow for errors on OceanInterface
3. Fixed disabled button for OceanInterface - Explorer
4. Created Section Title component

Fixes #290


app/components/SectionTitle.tsx
```diff
@@ -0,0 +1,16 @@
...
+import tailwind from 'tailwind-rn'
+
+export function SectionTitle ({ text, testID }: { text: string, testID: string }): JSX.Element {
+  return (
+    <Text
+      testID={testID}
+      style={[tailwind('p-4 text-xs text-gray-500 font-bold mt-2')]}
+    >
+      {
+        text
```
fuxingloh, line 12: minor suggestion: maybe you could have this as a {props.children} instead?

fuxingloh (approved):

ivan-zynesis (approved):

app/components/SectionTitle.tsx
thedoublejay: ah yeah good idea. If we have a section title with info symbol or tooltip, might be useful. I'll do it on next PR

PR #299, opened 2021-07-17
enable debug mode (temporary)

thedoublejay:

/kind fix

Temporary update for user testing, not going to be merged

Fixes #


app/screens/WalletNavigator/screens/WalletOnboarding.tsx
```diff
@@ -6,21 +6,20 @@ import { Mnemonic } from '../../../api/wallet/mnemonic'
 import { Text, View } from '../../../components'
 import { PrimaryColor, VectorIcon, VectorIconName } from '../../../constants/Theme'
 import { useWalletManagementContext } from '../../../contexts/WalletManagementContext'
-import { getEnvironment } from '../../../environment'
```
fuxingloh, line 9: might want to edit this file `debug: true` temporarily instead?

PR #328, opened 2021-07-23
Update expo-deps workflow to run every other even day

fuxingloh:

/kind chore


fuxingloh:
don't understand why is this workflow failing

thedoublejay:
@fuxingloh this won't work because `react-native-qrcode-svg` has a different RN dependency than Expo. `expo upgrade` is running on background, then it fails because `npm install` doesn't work. It needs to be `npm install --force`.

fuxingloh:
> react-native-qrcode-svg

We need to resolve it somehow, something for another day.

PR #343, opened 2021-07-25
Updated <Button> Component

thedoublejay:

/kind feature

Updated <Button> component. Accepts fill property (e.g, fill, outline, flat) or color (e.g, Primary, secondary)

Fixes #


app/components/Button.tsx
```diff
@@ -0,0 +1,34 @@
...
+import { ButtonProps, TouchableOpacity } from 'react-native'
+import { tailwind } from '../tailwind'
+import { Text } from './Text'
+
+export type ButtonFillTypes = 'fill' | 'outline' | 'flat'
+export type ButtonColor = 'primary' | 'secondary'
+
+export function Button (props: React.PropsWithChildren<ButtonProps> & { color?: ButtonColor, fill?: ButtonFillTypes, label?: string }): JSX.Element {
+  const buttonColor = props.color ?? 'primary'
+  const buttonFill = props.fill ?? 'fill'
```
fuxingloh, line 11:
```suggestion
  const { 
    color = 'primary',
    fill = 'fill',
    ...
  } = props
```

This kind of default restriction might be easier to grasps

app/components/Button.tsx
```diff
@@ -0,0 +1,34 @@
...
+                    ${buttonFill === 'fill' ? `bg-${buttonColor} bg-opacity-10` : 'bg-transparent'}`
+  const disabledStyle = 'bg-black bg-opacity-20 text-white text-opacity-5 border-0'
+
+  const textStyle = `${props.disabled === true ? 'text-white text-opacity-20' : `text-${buttonColor}`}`
+  return (
+    <TouchableOpacity
+      {...props}
+      testID={props.testID}
+      disabled={props.disabled}
+      onPress={props.onPress}
```
fuxingloh, line 22:
```suggestion
      {...props}
```

first line should be enough since it's redundant with `{...props`}?

fuxingloh (changes requested):

app/components/Button.tsx
thedoublejay: oh yeah haha good point

app/components/Button.tsx
```diff
@@ -0,0 +1,33 @@
+import React from 'react'
+import { ButtonProps, TouchableOpacity } from 'react-native'
+import { tailwind } from '../tailwind'
+import { Text } from './Text'
+
+export type ButtonFillTypes = 'fill' | 'outline' | 'flat'
+export type ButtonColor = 'primary' | 'secondary'
+
+export function Button (props: React.PropsWithChildren<ButtonProps> & { color?: ButtonColor, fill?: ButtonFillTypes, label?: string }): JSX.Element {
```
fuxingloh, line 9: Since this is an atomic component. I think it might be best to separate out the interface, contrary to my previous comment. Since this entire file serves a single purpose.

```suggestion
interface ButtonProps:  React.PropsWithChildren<ButtonProps> {
  color?: 'primary' | 'secondary'
  fill?: 'fill' | 'outline' | 'flat'
  label?: string
}

export function Button (props: ButtonProps): JSX.Element {
```

fuxingloh (changes requested):

fuxingloh (approved):

PR #365, opened 2021-07-28
`convert` - tokens vs UTXO screen

### LINKED ISSUE #272: `convert` - tokens vs UTXO screen
#### Why is this needed:
Screen to explain Tokens vs UTXO
image

kyleleow:

/kind feature

- Add tokens vs utxo screen that explains the differences

Fixes #272


app/screens/AppNavigator/screens/Balances/screens/TokensVsUtxoScreen.tsx
```diff
@@ -0,0 +1,61 @@
...
+import React from 'react'
+import { View } from 'react-native'
+import { Text } from '../../../../../components'
+import { translate } from '../../../../../translations'
+import { tailwind } from '../../../../../tailwind'
+import { getTokenIcon } from '../../../../../components/icons/tokens/_index'
+
+export function TokensVsUtxoScreen (): JSX.Element {
+  return (
+    <View testID='token_vs_utxo_screen'>
```
thedoublejay, line 11: Use ScrollView, also can you make background white, hard to read with gray color. Since we're using the same format on CreateWalletinfo page @jtlntno-cake

thedoublejay (changes requested):

thedoublejay:
Some comments:
1. Can we make the CTA of "UTXO vs Token" - use primary color? So it looks like a link. Currently, it looks plain since it's all grayed out. @jtlntno-cake 
2. Can you also change the icons on the Convert page? Using the proper icon for UTXO and Token. In the input and summary part @kyleleow

app/screens/AppNavigator/screens/Balances/screens/TokensVsUtxoScreen.tsx
kyleleow: Will change to ScrollView and white background.

I noticed flex-1 doesn't work on the child items of ScrollView to achieve the following two scenarios:
- fill out the empty space when screen's height > content's height
- able to scroll through content when screen's height < content's height

thedoublejay: Flex-1 is for width right? Also, ScrollView will render a wrapper div, which is why you need to have contentContainerStyle. 

No need to take full screen especially on mobile, just allow the content to have the normal layout

app/screens/AppNavigator/screens/Balances/ConvertScreen.tsx
```diff
@@ -193,11 +194,14 @@ function ToggleModeButton (props: { onPress: () => void }): JSX.Element {
   )
 }
 
-function TokenVsUtxosInfo (): JSX.Element {
+function TokenVsUtxosInfo (props: { navigation: NavigationProp<BalanceParamList>}): JSX.Element {
```
ivan-zynesis, line 197: navigation prop is in the context , use `useNavigation()` instead. When `navigation` passed around, it allow unpredictable user flow.

jtlntno-cake:
1. @thedoublejay Yes, let's do that for consistency. I updated the Marvel project to reflect this.

app/screens/AppNavigator/screens/Balances/screens/TokensVsUtxoScreen.tsx
```diff
@@ -0,0 +1,61 @@
...
+      <View style={tailwind('pt-6 px-4 border-b-2 border-gray-200')}>
+        <Text style={tailwind('text-black pb-6')}>
+          {translate('screens/ConvertScreen', 'DFI exists in two forms – UTXO and token, which are interchangeable depending on the usage. You can easily convert between the two forms with the app.')}
```
ivan-zynesis, line 14: `translate('screens/TokensVsUtxos', ...)`

should have its own 'section', esp this page has large amount of text required i18n

ivan-zynesis (changes requested):

app/screens/AppNavigator/screens/Balances/ConvertScreen.tsx
kyleleow: Thanks for the tips @ivan-zynesis! 
Could you also share some examples of the weird user flow so I can take note in the future?

thedoublejay (approved):

app/screens/AppNavigator/screens/Balances/ConvertScreen.tsx
ivan-zynesis: There are multiple types  of `navigation` prop. Our app now has `stack` then `bottom tab` as one of the page inside the stack. The component only inherit the access based on where itself located. eg a page inside the outer stack should not really navigate directly into a specific tab as the tab container has its on state to maintain.

ivan-zynesis (approved):

Issue #355, opened 2021-07-27
`receive` - e2e tests


thedoublejay:
#### What would you like to be added:
E2E tests for Receive Page
#### Why is this needed:

/triage accepted
/assign @izzycsy 
/area e2e

PR #366, opened 2021-07-29
e2e Test - `receive`

izzycsy:

/kind chore

• Added test for Balances/Receive

Fixes https://github.com/DeFiCh/wallet/issues/355


thedoublejay:
Need to add the following test scenarios:
1. Get the address generated in Receive Page
2. Add any amount and input the address on the Send page - Address field
3. It should trigger an address validation that will enable the submit button

thedoublejay (approved):

PR #351, opened 2021-07-27
`Encrypted Mnemonic Wallet`

ivan-zynesis:
#### What kind of PR is this?:

/kind feature

#### What this PR does / why we need it:
Create pin after verify mnemonic words
OceanInterface prompt pin interface

#### Which issue(s) does this PR fixes?:

Fixes #

#### Additional comments?:
Considerations of current implementation
1. Passcode prompt page UI accessible globally
2. OceanInterface remained for async task status display (after wallet unlocked) ie slow internet, broadcast take very long or fail.
3. OceanInterface remained as child of AppNavigator, accessible by tabbed pages for submit transaction (unlike prompt UI)
4. Thus all interactions with prompt UI actually done via WalletManagementContext

PR #403, opened 2021-08-03
`Encrypted Wallet` - onboard + transaction authorization

ivan-zynesis:

/kind feature

Encrypt wallet xpriv, no raw private key info at rest (even in secure store)

Fixes #

#### Additional comments?:
Move changes and replace #351

known glitches on mobile
- pin input component autofocus - do pop soft keyboard but not really focused on textfield (solved)
- activity indicator - spinner not showing (solved)
- encryption (nodeify scrypt js) implementation is slow - up to 10 seconds when wallet first created
- authorization page - cancel button not properly offset notification bar (added offset via padding but can be improved)


app/api/wallet/provider/mnemonic_encrypted.ts
```diff
@@ -15,17 +15,23 @@ const encryption = new PrivateKeyEncryption(new Scrypt(), numOfBytes => {
...
 
+interface PromptInterface {
+  prompt: PromptPassphrase
+}
+
 function initProvider (
   data: WalletPersistenceData<EncryptedProviderData>,
   network: EnvironmentNetwork,
-  promptPassphrase: PromptPassphrase
+  promptInterface: PromptInterface
```
ivan-zynesis, line 25: Hold the promise in another interface. The original implementation `PromptPassphrase` promise executed when this initProvider run, it should only execute when the prompt actually happening.

app/api/wallet/provider/mnemonic_encrypted.ts
```diff
@@ -15,17 +15,23 @@ const encryption = new PrivateKeyEncryption(new Scrypt(), numOfBytes => {
...
 ): EncryptedHdNodeProvider {
   if (data.type !== WalletType.MNEMONIC_ENCRYPTED || data.version !== 'v1') {
     throw new Error('Unexpected WalletPersistenceData')
   }
 
   const bip32Options = getBip32Option(network)
-  return EncryptedHdNodeProvider.init(data.raw, bip32Options, encryption, promptPassphrase)
+  return EncryptedHdNodeProvider.init(data.raw, bip32Options, encryption, async () => {
+    return await promptInterface.prompt()
```
ivan-zynesis, line 33: Hold the promise in another interface. The original implementation PromptPassphrase promise executed when this initProvider run, it should only execute when the prompt actually happening.

app/screens/WalletNavigator/WalletNavigator.tsx
```diff
@@ -19,6 +21,14 @@ export interface WalletParamList {
     words: string[]
   }
   RestoreMnemonicWallet: undefined
+  PinCreation: {
```
ivan-zynesis, line 24: onboarding flow
- show words (carry forward: words)
- verify words (carry forward: words, desired pin length)
- pin creation (carry forward: words, created pin)
- pin verification (encrypt and save wallet)

ivan-zynesis: this will be further extended later
4. pin verification (carry forward encrypted wallet data, verified pin)
5. enroll biometric (store wallet, store pin into secure store)

app/screens/RootNavigator.tsx
```diff
@@ -20,6 +21,7 @@ export function RootNavigator (): JSX.Element {
   return (
     <WalletProvider data={wallets[0]}>
       <WalletAddressProvider>
+        <TransactionAuthorization />
```
thedoublejay, line 24: This one shouldn't be imported if the wallet is not encrypted. The problem is it's causing the UI to break if the current loaded wallet is an unencypted one. 

For iOS, even if you uninstall the app, the localStorage data is still there. Previous users who tested the unencrypted wallet will not be able to test since they will have infinite loop of encypted error.

thedoublejay (changes requested):

app/components/Button.tsx
```diff
@@ -8,6 +8,8 @@ interface ButtonProps extends React.PropsWithChildren<Props> {
   fill?: 'fill' | 'outline' | 'flat'
   label?: string
   margin?: string
+  delayLongPress?: number
+  onLongPress?: () => void
```
fuxingloh, line 12: This prop should be refactored to extends`React.PropsWithChildren<TouchableOpacityProps>`, to open an issue to resolve in another PR

app/screens/RootNavigator.tsx
fuxingloh: This component should be nested within `MnemonicEncryptedProvider`

https://github.com/DeFiCh/wallet/pull/403/files#diff-6401c3d7908d3b27690269213369a1ad75c5b40f42260e28db3a515b14dc6e59R100

app/contexts/WalletContext.tsx
```diff
@@ -77,31 +79,24 @@ function MnemonicUnprotectedProvider (props: WalletProviderProps<MnemonicProvide
...
+  const encryptedWalletInterface: EncryptedWalletUIContext = {
+    provide: (ewi: EncryptedWalletInterface) => {
+      setPromptUI(ewi)
+    }
+  }
 
   return (
     <WalletContext.Provider value={wallet}>
-      <EncryptedWalletContext.Provider value={signingRef}>
+      <EncryptedWalletContext.Provider value={encryptedWalletInterface}>
```
fuxingloh, line 99:
```suggestion
      <EncryptedWalletContext.Provider value={encryptedWalletInterface}>
        <TransactionAuthorization />
```

app/contexts/WalletContext.tsx
```diff
@@ -13,24 +14,25 @@ import { useNetworkContext } from './NetworkContext'
...
 
 const WalletContext = createContext<WhaleWallet>(undefined as any)
-const EncryptedWalletContext = createContext<MutableRefObject<EncryptedWalletInterface>>(undefined as any)
+const EncryptedWalletContext = createContext<EncryptedWalletUIContext>(undefined as any)
 
 export function useWallet (): WhaleWallet {
   return useContext(WalletContext)
 }
 
+export function useEncryptedWallet (): EncryptedWalletUIContext {
```
fuxingloh, line 32:
```suggestion
export function useEncryptedWalletUI (): EncryptedWalletUIContext {
```

fuxingloh (changes requested):
Great implementation! with use of `const wallet = useMemo(() => {` and recreating wallet when the prompt is setup.

app/contexts/WalletContext.tsx
ivan-zynesis: Argh this make sense but implementation is impossible. The <TransactionAuthorization /> here will mess up the playground page UI in browser

app/screens/RootNavigator.tsx
ivan-zynesis: @thedoublejay technically it is a dumb component.

From EncryptedWalletProvider perspective, this is a prompting UI to collect a 6 digits string.

From TransactionAuth component perspective
```
// const encryptionUI = useEncryptedWalletUI() // this is not really this component concern
const { pinConsumer } = props
pinConsumer.provide(this) // myself

onPrompt () => {
  // wait for user
  return userInput // doesnt matter what prompt consumer do this with
}
```

ivan-zynesis: > For iOS, even if you uninstall the app, the localStorage data is still there. Previous users who tested the unencrypted wallet will not be able to test since they will have infinite loop of encypted error.

Wallet should be wiped after 4 attempts and see onboarding flow again.
Hmm this is more like UX to solve issue. If an iphone ownership somehow transferred from one light wallet user to another. (later there should be implementation: when app focus opened/focused, prompt for passphrase) Maybe there is an option "Not my wallet, wipe and re-onboard"

thedoublejay: @ivan-zynesis it's not even related to components. There's an error thrown on the useEffect. It's due to having an unencrypted wallet loaded

app/components/Button.tsx
fuxingloh: @thedoublejay open an issue for this

fuxingloh (approved):

thedoublejay (approved):

PR #415, opened 2021-08-04
update address url

thedoublejay:

/kind chore

Fixes #


fuxingloh (approved):

fuxingloh:
Some of the test operations where we query `cypress` via `playground` can be simplified to just query whale or playground. I think we should have an issue to refactor that in the future. Low priority.

PR #433, opened 2021-08-05
factor in slippage for poolswap calculation

### LINKED ISSUE #341: review `poolswap` calculations
#### What happened:
Revisit calculations for poolswap. Currently, the values are less than the received amount. In the desktop app, we're using `testpoolswap` for the update prices. Might be different in the mobile wallet but worth checking.

thedoublejay:

/kind fix

Factor in slippage for pool swap to have more accurate results. Updated e2e to check if the poolswap amount is same

Fixes #341


canonbrother (approved):
so the ux is following testpoolswap

fuxingloh (approved):
> UX wise we should also display something about slippage and worst-case scenario
> U-Zyn

@thedoublejay to create an issue to track the above

PR #451, opened 2021-08-06
`global state` - when switching to Mainnet, global state is persisted

### LINKED ISSUE #443: `global state` - when switching to Mainnet, global state is persisted
#### What happened:
When switching network to mainnet, tokens from playground are persisted.

E.g, 
DFI UTXO and Token are 0 but BTC Playground is still there.

1. Create a playground wallet
2. Top up tokens
3. Change network to mainnet
4. Create a new wallet
5. Tokens are still there

What we need to do is everytime you Exit a wallet, you should reset global state as well.
/triage accepted

kyleleow:

/kind fix

- Replace `wallet.get(0).getAddress()` with `useWalletAddressContext()`
- In order to use the correct address when calling transaction list API after switching to MainNet

Fixes #443

#### Additional comments?:
- Balance is not updated because token API is returning error
- Pending update from @fuxingloh to check on token API for further investigation


fuxingloh:
My recent commit, adds `WalletStoreProvider` that is memoized to network & wallets setting. 

There was a unique store in Playground & Wallet. There should only be one store, this refactor hoists the store up one hierarchy level.

thedoublejay (approved):

Issue #466, opened 2021-08-07
display `active network`


thedoublejay:
#### What UI/UX improvements would you like to be implemented: 
Display active network to prevent user doing transactions on other networks.

PR #484, opened 2021-08-07
refactor & fixes for `Network` & `Wallet` issues, `DEX` - APR, `Header` - Network

### LINKED ISSUE #423: `DEX` - show APR on pool pairs
Show APR for DEX pool pairs
#### Why is this needed:
/area ui-ux wallet
/priority important-soon
/triage accepted

### LINKED ISSUE #470: `network` not persisted on app re-open
#### What happened:
The current active network is not persisted when you re-open the app. Example, switch to Mainnet and re open the app. It reverts to Playground.
/triage accepted

### LINKED ISSUE #471: `address` - on change network, remains the same
#### What happened:
On change of network, address remains the same.

### LINKED ISSUE #472: `wallet` - created wallet for mainnet doesn't load
#### What happened:
When you have a mainnet and playground wallet loaded, and you exit playground wallet. It will go to onboarding page, which is expected. When you re-open the app, it should load mainnet wallet since it's existing.

### LINKED ISSUE #473: `onboarding` - restoring a wallet using mainnet mnemonic
#### What happened:
Restoring a wallet using mainnet mnemonic, loads playground network

### LINKED ISSUE #480: `network` - switching networks only update state once
#### What happened:
On initial network change, state updates. However on reverting back to original network. State doesn't change anymore.
Playground -> mainnet (state updates)
Mainnet -> playground (state doesn't change anymore)

fuxingloh:
/kind fix
/kind refactor

Fixed #470 by removing `useConnectedPlayground()` which was deprecated, it auto-resolves the network on start by looking for a playground network to connect to. This implementation is just frustrating as it doesn't have a deterministic or static behavior to network resolution. DeFi Scan has removed this implementation recently as it was difficult to manage SSR and SPA network states, the resolution on DeFi Scan is now static too.
- This also fixes #473
- By removing `useConnectedPlayground()` playground local/remote resolution is now not automated anymore. By default, it will always resolve to the first network (which is "Local" for CI and development). During development, you must switch networks manually. And for e2e testing you must always use Local Network, you can start local network via `npm run playground:start`.
- To allow network switching easily during CI and testing, we can add it into "DeFi Playground", now "DeFi Testing".
- `PlaygroundScreen` has been refactored from "DeFi Playground" to "DeFi Testing" this refactor in large is to allow other networks to use the new "DeFi Testing" environment for features such as quick Wallet setup. A large part of the code into DeFi Playground has been refactored to be componentized. 

Fixed #471 by adding missing dependencies to `useMemo` in WalletAddressToken: `}, [wallet])`

Fixed #472 by deploying `whale:0.8.6`

Fixed #480 by adding a few missing dependencies to `useEffect`
- `}, [address])`
- `}, [wallet])`

Linked PRs: https://github.com/DeFiCh/wallet/pull/488 https://github.com/DeFiCh/wallet/pull/490

Fixed #423 #466

> I only tested these on `Web` and some parts of it on `Mobile`. @thedoublejay and someone else need to verify them

#### Additional comments?:

~~Many `useEffect` uses `}, [JSON.stringify(tokens)])` we should refactor this. `JSON.stringy()` shouldn't be a thing for `useEffect`. Many `useEffect` dependencies array is not well thought out, we need to be inspect them carefully.~~

Our context and provider are deeply nested which each resolution relying on a previous upstream provider/context this means if network changes whale will be updated, if whale is updated so must wallet, starting from: 
1. network
2. playground 
3. whale
4. wallet persistence 
5. defi scan
6. wallet store
7. wallet 
8. wallet address
9. transactions

```html
<ErrorBoundary>
  <NetworkProvider>
    <PlaygroundProvider>
      <WhaleProvider>
        <WalletPersistenceProvider>
          <DeFiScanProvider>
            <WalletStoreProvider>
              <SafeAreaProvider>
                <WalletNavigator />
                or
                <WalletProvider data={wallets[0]}>
                  <WalletAddressProvider>
                    <TransactionAuthorization />
                    <NavigationContainer>
                      <App.Navigator>
                        <App.Screen name='App' component={BottomTabNavigator} />
                        <App.Screen name='Playground' component={PlaygroundNavigator} />
                      </App.Navigator>
                    </NavigationContainer>
                  </WalletAddressProvider>
                </WalletProvider>
              </SafeAreaProvider>
            </WalletStoreProvider>
          </DeFiScanProvider>
        </WalletPersistenceProvider>
      </WhaleProvider>
    </PlaygroundProvider>
  </NetworkProvider>
</ErrorBoundary>
```

### DIFF
--- a/.idea/dictionaries/fuxing.xml
+++ b/.idea/dictionaries/fuxing.xml
@@ -173,6 +173,7 @@
       <w>scripthash</w>
       <w>secp</w>
       <w>segwit</w>
+      <w>semibold</w>
       <w>sendrawtransaction</w>
       <w>sendtoaddress</w>
       <w>sendtokenstoaddress</w>
--- a/App.tsx
+++ b/App.tsx
@@ -4,7 +4,7 @@ import './_shim'
 import { Logging } from './app/api'
 import { DeFiScanProvider } from './app/contexts/DeFiScanContext'
 import { NetworkProvider } from './app/contexts/NetworkContext'
-import { PlaygroundProvider, useConnectedPlayground } from './app/contexts/PlaygroundContext'
+import { PlaygroundProvider } from './app/contexts/PlaygroundContext'
 import { WalletPersistenceProvider } from './app/contexts/WalletPersistenceContext'
 import { WalletStoreProvider } from './app/contexts/WalletStoreProvider'
 import { WhaleProvider } from './app/contexts/WhaleContext'
@@ -21,16 +21,9 @@ initI18n()
  * - CachedPlaygroundClient
  */
 export default function App (): JSX.Element | null {
-  const isLoaded: boolean[] = [
-    useCachedResources(),
-    // find a connected playground at app load
-    // TODO(fuxingloh): feel like we should deprecate to auto resolve to a fixed network
-    //  instead of automated resolution, after setup we switch to one based on the test?
-    //  might be difficult due to ci automation?
-    useConnectedPlayground()
-  ]
+  const isLoaded = useCachedResources()
 
-  if (isLoaded.includes(false)) {
+  if (!isLoaded) {
     SplashScreen.preventAutoHideAsync()
       .catch(Logging.error)
     return null
--- a/app/components/HeaderTitle.tsx
+++ b/app/components/HeaderTitle.tsx
@@ -1,0 +1,71 @@
+import React, { useEffect } from 'react'
+import { View } from 'react-native'
+import { useDispatch, useSelector } from 'react-redux'
+import { Logging } from '../api'
+import { useNetworkContext } from '../contexts/NetworkContext'
+import { useWhaleApiClient } from '../contexts/WhaleContext'
+import { RootState } from '../store'
+import { block } from '../store/block'
+import { tailwind } from '../tailwind'
+import { Text } from './Text'
+
+export function ConnectionStatus (): JSX.Element {
+  const { network } = useNetworkContext()
+  const connected = useSelector((state: RootState) => state.block.connected)
+  const isPolling = useSelector((state: RootState) => state.block.isPolling)
+  const api = useWhaleApiClient()
+
+  const dispatch = useDispatch()
+
+  useEffect(() => {
+    let interval: number
+
+    function refresh (): void {
+      dispatch(block.actions.setPolling(true))
+      api.stats.get().then(({ count }) => {
+        dispatch(block.actions.updateBlock({ count: count.blocks }))
+        dispatch(block.actions.setConnected(true))
+      }).catch((err) => {
+        dispatch(block.actions.updateBlock({ count: 0 }))
+        dispatch(block.actions.setConnected(false))
+        Logging.error(err)
+      })
+    }
+
+    if (!isPolling) {
+      refresh()
+      interval = setInterval(refresh, 3000)
+    }
+    return () => {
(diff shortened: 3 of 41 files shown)


thedoublejay:
@fuxingloh regarding the json.stringify dependency. I added it because it needs to be deep equality, like balance for example. It was actually recommended by one of the react mainteners. For ref https://stackoverflow.com/questions/59467758/passing-array-to-useeffect-dependency-list

fuxingloh:
> @fuxingloh regarding the json.stringify dependency. I added it because it needs to be deep equality, like balance for example. It was actually recommended by one of the react mainteners. For ref https://stackoverflow.com/questions/59467758/passing-array-to-useeffect-dependency-list

Oh, didn't know this. Probably worth a refactor to unify our `useTokens()` implementation to be deeply reactive in the store. Probably wasn't a good idea to put dex state outside of store.

thedoublejay:
@fuxingloh I'm adding some tests for this, I'll create a PR to your branch to test all these bugs.

thedoublejay:
@fuxingloh  There's still an issue on the address side. 

Main problem:
1. Create a wallet in Local Network
2. Send some tokens, poolpairs etc.
2. Change network to Mainnet
3. Change back to Local
4. Tokens, poolpairs are not displayed.

The issue is on the initial wallet creation and the network switch, there are two different address created for the same Local Wallet. Since address is a parameter for all of our APIs, it doesn't return the correct value.

fuxingloh:
> @fuxingloh
> Address is not the same on network change
> Main problem:
> 
> 1. Create a wallet in Local Network
> 2. Send some tokens, poolpairs etc.
> 3. Change network to Mainnet
> 4. Change back to Local
> 5. Tokens, poolpairs are not displayed.
> 
> The issue is on the initial wallet creation and the network switch, there are two different address created for the same Local Wallet. Since address is a parameter for all of our APIs, it doesn't return the correct value. Because of this, all user related data doesn't return correctly (e.g, PoolPairs, Balances etc.)
> 
> This PR - #488 has e2e tests that will pass if these bugs are fixed.

I will debug this now

fuxingloh:
> > @fuxingloh
> > Address is not the same on network change
> > Main problem:
> > 
> > 1. Create a wallet in Local Network
> > 2. Send some tokens, poolpairs etc.
> > 3. Change network to Mainnet
> > 4. Change back to Local
> > 5. Tokens, poolpairs are not displayed.
> > 
> > The issue is on the initial wallet creation and the network switch, there are two different address created for the same Local Wallet. Since address is a parameter for all of our APIs, it doesn't return the correct value. Because of this, all user related data doesn't return correctly (e.g, PoolPairs, Balances etc.)
> > This PR - #488 has e2e tests that will pass if these bugs are fixed.
> 
> I will debug this now

The latest PR might fix it,

1. Added `address` to a 2 useEffect that might need it
2. `WalletStoreProvider` to only refresh on `wallets` deps since wallets depends on network
3. **Should be the root cause**: Added missing `dispatch` deps to `useTokenAPI` because store is now `useMemo()` and that causes store state to refresh, it's the last the refresh

thedoublejay:
@fuxingloh I think the issue is the getWallet(0) address returns a different one. I saw it on the WalletAddressProvider, when setting on address it's a different address

thedoublejay (approved):
Nice fixes

fuxingloh:
The dependencies array is just wild west now lol, we need to tackle them soon.

PR #506, opened 2021-08-09
Add explicit label for Token DFI

### LINKED ISSUE #447: explicit labels for UTXO and Token DFI
#### What happened:
“Add Liquidty” page, it should be shown as DFI (Token). The example here, I have DFI, but no token form. Took me a bit to realize that it’s because I’m missing token.

kyleleow:

/kind fix

- To allow user to differentiate between DFI (UTXO) and DFI (Token)

Fixes #447


thedoublejay:
@kyleleow can you check pool swap page also?

thedoublejay (approved):

PR #458, opened 2021-08-06
`Settings` - change passcode

ivan-zynesis:
#### What kind of PR is this?:

/kind feature

#### What this PR does / why we need it:
More security settings implementation

#### Which issue(s) does this PR fixes?:

Fixes #

PR #526, opened 2021-08-11
`transaction ui` - fixes

### LINKED ISSUE #462: `transaction authorization` - cancel button
#### What happened:
Clicking on cancel button, intermittently loads the screen again. On the code, once onCancel is called, it calls onPrompt that triggers emitEvent('pin') event. Test case for cancel is failing due to this issue, you can refer to this - https://github.com/DeFiCh/wallet/blob/68613513c426a07d80636c5f06390a4d4f857ca9/cypress/integration/functional/wallet/authorization/transactionAuthorization.spec.ts#L14

/triage accepted
/assign @ivan-zynesis 
/area e2e

### LINKED ISSUE #475: `transaction authorization` - cancel should not reset attempts
#### What happened:

- [ ] Cancelling the transaction prompt resets the attempts. It might be a root cause of brute force attack on pin code as they will get unlimited attempts. Both on transaction and non-transaction prompts.

- [ ] Another issue on transaction prompts, on cancel the error is still displayed there. Does not happen on non transaction ui

### LINKED ISSUE #513: `transaction authorization` - on open, automatic validation
#### What happened:
When signing a transaction, the transaction authorization automatically triggers a validation causing it to close.

https://user-images.githubusercontent.com/25013382/128797650-ec6c73c4-fe6f-40f8-ba54-3e87c338b42b.mp4

thedoublejay:

/kind fix

1. Fix issues with Transaction UI
2. Fixes #513 - It happens because there's an error on the initial transaction and not on the signing itself. Causing it to auto-retry even without any inputs
3. Fixes #475 - Attempt is recovered from LocalStorage
4. Fixes #462 - Cancelling should work now and not re-trigger
5. Added e2e tests to cover all these scenarios

Fixes #

### DIFF
--- a/app/screens/TransactionAuthorization.tsx
+++ b/app/screens/TransactionAuthorization.tsx
@@ -172,27 +172,26 @@ export function TransactionAuthorization (): JSX.Element | null {
             onSubmitCompletion()
           })
       } else if (authentication !== undefined) {
-        let status = ''
+        let invalidPassphrase = false
         authenticateFor(onPrompt, authentication, onRetry)
+          .then(async () => {
+            setAttemptsRemaining(MAX_PASSCODE_ATTEMPT)
+            await PasscodeAttemptCounter.set(0)
+          })
           .catch(e => {
             // error type check
             if (e.message === INVALID_HASH) {
-              status = INVALID_HASH
-            } else if (e.message === USER_CANCELED) {
-              status = USER_CANCELED
+              invalidPassphrase = true
             } else if (e.message !== USER_CANCELED) {
               dispatch(ocean.actions.setError(e))
             }
           })
           .then(async () => {
-            if (status === INVALID_HASH) {
+            if (invalidPassphrase) {
               setAttemptsRemaining(MAX_PASSCODE_ATTEMPT)
               await PasscodeAttemptCounter.set(0)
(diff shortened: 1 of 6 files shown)


app/screens/TransactionAuthorization.tsx
```diff
@@ -120,49 +142,63 @@ export function TransactionAuthorization (): JSX.Element | null {
...
           })
           .then(async () => {
-            if (invalidPassphrase) {
+            if (status === INVALID_HASH) {
+              setAttemptsRemaining(MAX_PASSCODE_ATTEMPT)
+              await PasscodeAttemptCounter.set(0)
               await clearWallets()
               onUnlinkWallet()
-            } else {
+            } else if (status !== USER_CANCELED) {
```
ivan-zynesis, line 193: please use the fix in #458 (which is more intuitive, do not have to wait for error preprocess for valid passcode callback)

thedoublejay: ah yeah make sense, let me try it locally first

thedoublejay: Updated @ivan-zynesis

ivan-zynesis (approved):

PR #527, opened 2021-08-11
Fix passcode prompt cancelation reset attempt counter

ivan-zynesis:

/kind fix

bug: if (not invalid code) reset // can be any other error
fix: if (valid) reset

Fixes #


thedoublejay:
@ivan-zynesis  I have fixed it here - https://github.com/DeFiCh/wallet/pull/526

PR #528, opened 2021-08-11
`balance` - add receive and share button on balances header

### LINKED ISSUE #464: share my public address
Easy access button that triggers the native share UI of Android/iOS.

### LINKED ISSUE #524: Receive `[Button]` should be at /Balances

There should be a receive [Button] at /Balances page.

#### Why is this needed:

Users might want to receive a `DST token`, but they don't have the token yet hence there isn't an option for it. 

They could open the `DFI UTXO` or `DFI Token` where there is a receive option, however, this isn't obvious for your average user.

Balances Tab

/triage accepted
/priority important-soon

thedoublejay:

/kind feature

Fixes #524 
Fixes #464 

#### Additional comments?:
image


app/screens/AppNavigator/screens/Balances/BalancesNavigator.tsx
```diff
@@ -27,17 +31,70 @@ export interface BalanceParamList {
...
         component={BalancesScreen}
         options={{
+          headerTitleAlign: 'center',
           headerTitle: () => <HeaderTitle text={translate('screens/BalancesScreen', 'Balances')} />,
-          headerBackTitleVisible: false
+          headerBackTitleVisible: false,
+          headerLeft: () => (
+            <BalanceActionButton
+              testID='header_share_balance' icon='share'
+              onPress={async () => await onShare(address)}
```
fuxingloh, line 89: Best to have share embedded within `Receive` screen, contextual you are sharing Address. While having share on this screen means you are sharing your balances.

fuxingloh (changes requested):

app/screens/AppNavigator/screens/Balances/BalancesNavigator.tsx
```diff
@@ -27,17 +29,53 @@ export interface BalanceParamList {
...
 
 export function BalancesNavigator (): JSX.Element {
+  const navigation = useNavigation()
   return (
     <BalanceStack.Navigator>
       <BalanceStack.Screen
         name='Balances'
         component={BalancesScreen}
         options={{
+          headerTitleAlign: 'center',
```
fuxingloh, line 70:
```suggestion
```

don't think we should center

fuxingloh (approved):

app/screens/AppNavigator/screens/Balances/BalancesNavigator.tsx
thedoublejay: I'll add this on another PR

PR #533, opened 2021-08-11
Fix unable to resolve dftx signing promise

### LINKED ISSUE #522: `transaction authorization` - After entering the passcode to sign a transaction, stuck at that page and transaction doesn't get signed.
#### What happened:

On iOS TestFlight v0.3.2, Playground

1. Send 0.5 `USDT` to another address
1. After entering the passcode to sign a transaction, stuck at that page and the transaction doesn't get signed.

ivan-zynesis:

/kind fix

Fixes #522


app/screens/TransactionAuthorization.tsx
```diff
@@ -103,21 +100,19 @@ export function TransactionAuthorization (): JSX.Element | null {
...
+      return await new Promise<string>((resolve, reject) => {
+        // passphrase prompt is meant for authorizing single transaction regardless
+        // caller should not prompt for next transaction before one is completed
+        // proxy the promise, wait for user input
+        PASSPHRASE_PROMISE_PROXY = {
+          resolve, reject
         }
-      }
-    })
+        emitEvent('PIN')
```
ivan-zynesis, line 113: Better state management: only update the state if the promise proxied.

app/screens/TransactionAuthorization.tsx
```diff
@@ -282,7 +278,7 @@ export function TransactionAuthorization (): JSX.Element | null {
           message={status === 'SIGNING' ? translate('screens/TransactionAuthorization', 'Signing...') : undefined}
         />
         {
-          (isSubmitted && attemptsRemaining !== undefined && attemptsRemaining !== MAX_PASSCODE_ATTEMPT) ? (
+          (attemptsRemaining !== undefined && attemptsRemaining !== MAX_PASSCODE_ATTEMPT) ? (
```
ivan-zynesis, line 281: I think the intention here is to show the error only after the at least one trial. However, the reminder should be there even no "new" attempt.

eg
Reach prompt UI, fail 3 times, cancel.
Try another transaction, user should see the reminder only 1 chance left, another failure lead to wallet backup wipe.

thedoublejay: Yep correct, but if that's the case, then we need to update the copy for it. Currently it has 'Incorrect Input' even without any inputs.

ivan-zynesis: Will check again, haven't seen that message.

app/screens/TransactionAuthorization.tsx
```diff
@@ -309,7 +305,6 @@ async function execWithAutoRetries (promptPromise: () => Promise<any>, onAutoRet
     return await promptPromise()
   } catch (e) {
     Logging.error(e)
-    retries = await PasscodeAttemptCounter.get() ?? 0
```
ivan-zynesis, line 312: Remove persistent storage access from here.

1. SecureStore is a relatively expensive/heavy resource to be accessed (quite slow), submit via state is much more efficient
2. trying to minimize the risk if any uncaught error here, potentially lead to unhandled statement management

app/screens/TransactionAuthorization.tsx
```diff
@@ -281,14 +281,22 @@ export function TransactionAuthorization (): JSX.Element | null {
...
+          (isRetry && attemptsRemaining !== undefined && attemptsRemaining !== MAX_PASSCODE_ATTEMPT) ? (
             <Text testID='pin_attempt_error' style={tailwind('text-center text-error text-sm font-bold mt-5')}>
               {translate('screens/PinConfirmation', `${attemptsRemaining === 1 ? 'Last attempt or your wallet will be unlinked for your security'
                 : 'Incorrect passcode. %{attemptsRemaining} attempts remaining'}`, { attemptsRemaining: `${attemptsRemaining}` })}
             </Text>
           ) : null
         }
+        {// on first time: warn user there were accumulated error attempt counter
+          (!isRetry && attemptsRemaining !== undefined && attemptsRemaining !== MAX_PASSCODE_ATTEMPT) ? (
+            <Text testID='pin_attempt_warning' style={tailwind('text-center text-error text-sm font-bold mt-5')}>
```
ivan-zynesis, line 294: Add as separate text component, easier to change copywriting and logic.

app/screens/TransactionAuthorization.tsx
```diff
@@ -54,38 +54,32 @@ export function TransactionAuthorization (): JSX.Element | null {
...
-    emitEvent('CANCELLED')
-    setIsSubmitted(false)
     if (PASSPHRASE_PROMISE_PROXY !== undefined) {
-      const reject = PASSPHRASE_PROMISE_PROXY.reject
-      setTimeout(() => {
-        reject(new Error(USER_CANCELED))
-        // remove proxied promised, allow next prompt() call
-        PASSPHRASE_PROMISE_PROXY = undefined
-      }, 50)
+      PASSPHRASE_PROMISE_PROXY.reject(new Error(USER_CANCELED))
```
ivan-zynesis, line 75: do not require extra setState here, rejecting the pin promise with USER_CANCELED error lead the signing/authenticate result handler.

thedoublejay (approved):

thedoublejay (approved):

PR #546, opened 2021-08-12
Fix unable to sign tx

### LINKED ISSUE #522: `transaction authorization` - After entering the passcode to sign a transaction, stuck at that page and transaction doesn't get signed.

### LINKED ISSUE #541: `transaction authorization` - on open, attempt error message stutters
#### What happened:
When user has an existing attempt and cancelled, re-opening it causes some stutter of the error message. It displays as 2 attempts then changes to last attempt

https://user-images.githubusercontent.com/25013382/129121412-4b442b4b-db43-4089-aa25-fc332bad2fc9.MP4

ivan-zynesis:

/kind fix

cyclic hook dep causing passcode prompting fail state check

Fixes #522 
Fixes #541

### DIFF
--- a/app/screens/TransactionAuthorization.tsx
+++ b/app/screens/TransactionAuthorization.tsx
@@ -1,6 +1,6 @@
 import { CTransactionSegWit } from '@defichain/jellyfish-transaction/dist'
 import { WhaleWalletAccount } from '@defichain/whale-api-wallet'
-import React, { Dispatch, useCallback, useEffect, useState } from 'react'
+import React, { useCallback, useEffect, useState } from 'react'
 import { ActivityIndicator, Alert, Platform, SafeAreaView, TouchableOpacity } from 'react-native'
 import { useDispatch, useSelector } from 'react-redux'
 import { Logging } from '../api'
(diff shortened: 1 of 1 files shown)


app/screens/TransactionAuthorization.tsx
```diff
@@ -80,43 +80,32 @@ export function TransactionAuthorization (): JSX.Element | null {
...
-        // passphrase prompt is meant for authorizing single transaction regardless
-        // caller should not prompt for next transaction before one is completed
-        // proxy the promise, wait for user input
-        PASSPHRASE_PROMISE_PROXY = {
-          resolve, reject
-        }
-        emitEvent('PIN')
-      })
-    }
-  }, [status])
```
ivan-zynesis, line 114: main issue, although this callback is aware (dep list) of `status` update. However, the instance of `onPrompt` passed/provided do use the initial value.

Adding `onPrompt` as dep list to useEffect will lead to cyclic dep child->parent->child->... infinite re-render.
Have to give up the extra check here.

The status check before start signing/authenticate is the guard to prevent start prompt for pin.

app/screens/TransactionAuthorization.tsx
```diff
@@ -80,43 +80,32 @@ export function TransactionAuthorization (): JSX.Element | null {
...
   const onRetry = useCallback(async (attempts: number) => {
     setIsRetry(true)
-    setPin('')
     setAttemptsRemaining(MAX_PASSCODE_ATTEMPT - attempts)
     await PasscodeAttemptCounter.set(attempts)
-    emitEvent('PIN')
   }, [attemptsRemaining])
 
-  // transaction signing specific callback
-  const onComplete = useCallback(async (dispatch: Dispatch<any>, tx: CTransactionSegWit) => {
```
ivan-zynesis, line 90: move this single use callback declaration straight into signTx.then

app/screens/TransactionAuthorization.tsx
```diff
@@ -80,43 +80,32 @@ export function TransactionAuthorization (): JSX.Element | null {
...
-      })
-    }
-  }, [status])
-
-  const onSubmitCompletion = (): void => {
+  const onTaskCompletion = (): void => {
     setPin('')
-    emitEvent('IDLE')
     setIsRetry(false)
+    emitEvent('IDLE') // very last step, open up for next task
```
ivan-zynesis, line 108: the one and only possible IDLE event emitter (except initialization), as the very last step in the `finally` block

thedoublejay (approved):

PR #575, opened 2021-08-13
`pool swap` - confirm page

### LINKED ISSUE #552: `poolswap` - confirm page
For all transactions, we need to have a confirm page for consistency. Please check the e2e tests as well for regressions.

image

thedoublejay:

/kind feature

1. Pool swap confirm page

Fixes #552


ivan-zynesis (approved):
This PR looks good to me.

I think I would like to improve the `postAction` to `postAction(error? Error)`, and it should be only emitted in finally block to ensure the flow go correctly, we can do that next week

thedoublejay:
@ivan-zynesis Finally block will be called regardless if it's an error or not. Also, it will take a long time to confirm so the postAction won't take effect immediately. My implementation is to call it after broadcast so it won't take long

PR #577, opened 2021-08-13
fix for swap price

### LINKED ISSUE #578: `pool swap` - on swap of tokens, price is not updated
#### What happened:
On swap of tokens, price is not updated causing invalid price calculation

thedoublejay:

/kind fix

1. When swapping tokens, pool swap price remains

Fixes #578


fuxingloh (approved):
try to avoid multiline ternary, easier to understand with `if()`

PR #614, opened 2021-08-16
refactor to `npm run lint` instead of `npm run standard`

fuxingloh:

/kind refactor

Same lint style but using `eslint` runner by default for better IDE integration and customizable to expo defaults.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,0 +1,19 @@
+{
+  "parserOptions": {
+    "project": "./tsconfig.json"
+  },
+  "extends": [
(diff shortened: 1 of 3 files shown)


thedoublejay:
@fuxingloh can we add prettier? 🙏

thedoublejay (approved):

fuxingloh:
> @fuxingloh can we add prettier? 🙏

😅 I hate prettier, it's opinionated, Creates too many lines of code when easily expressable with just one.

PR #625, opened 2021-08-16
fix for poolswap

thedoublejay:

/kind fix

1. Added disable for Token B input as we don't allow updating it yet
2. Added params on cancel

Fixes #


app/screens/AppNavigator/screens/Dex/PoolSwap/PoolSwapScreen.tsx
```diff
@@ -85,7 +85,8 @@ export function PoolSwapScreen ({ route }: Props): JSX.Element {
         tokenA,
         tokenB,
         swap,
-        fee
+        fee,
+        poolpair
```
ivan-zynesis, line 89: where is this used?

thedoublejay: when cancelling, need to pass params again.

thedoublejay: https://github.com/DeFiCh/wallet/blob/141c024490291b2fda3979e39c5bfe4dcd260869/app/screens/AppNavigator/screens/Dex/PoolSwap/ConfirmPoolSwapScreen.tsx#L65

fuxingloh:
image
 

should reset value when we click that button

thedoublejay:
Updated @fuxingloh

fuxingloh:
1 last thing is this should be estimated too:

image

thedoublejay:
@fuxingloh updated title

fuxingloh (approved):

PR #626, opened 2021-08-16
`Alert` - Handling on Native and Web

### LINKED ISSUE #389: `Alert` - Handling on Native and Web
Currently, `Alert` component is only supported on Native devices. For web, it doesn't display. This might cause some inconsistent behavior on e2e testing. 
#### Why is this needed:
Add a component that handles Alerts for web and native. So during e2e, it can have consistent behavior. Example, when verifying recovery words, there's an Alert on mobile on error. Clicking on the alert will redirect to previous page. On web, we don't have this behavior, it will only redirect to previous page without error message.

image

/triage accepted
/area ui-ux

### LINKED ISSUE #567: `create wallet` - on click back
When you're creating a wallet, and go back to the previous screen. Alll the mnemonic words will reset. Add a warning to the user that when he exits the mnemonic page, it will reset all current words.

image

kyleleow:

/kind feature

1. To improve e2e testing that involves react native `Alert.alert()` on web
2. Add warning on create wallet back

Fixes #389
Fixes #567 

#### Additional comments?:


app/components/CustomAlert.tsx
```diff
@@ -0,0 +1,46 @@
...
+ * - does not support custom button text
+ * - single button will always has two buttons, in order to implement `onPress` callback
+ *
+ * ## Native
+ *
+ * Using react-native's `Alert`.
+ *
+ * @reference https://github.com/necolas/react-native-web/issues/1026#issuecomment-687572134
+ */
+export function CustomAlert (option: CustomAlertOption): void {
```
thedoublejay, line 30:
```suggestion
export function WalletAlert (option: CustomAlertOption): void {
```

thedoublejay (changes requested):

thedoublejay (approved):

thedoublejay (approved):

PR #628, opened 2021-08-17
Customizable `TransactionAuthorization` labels

### LINKED ISSUE #496: Reveal Recovery Pin Code copy is same as signing transaction, it should be different

The copy in the reveal recovery pin code is the same as the signing transaction copy. We need an alternative copy for that.

/triaged accepted
/area ui-ux

### LINKED ISSUE #560: `transaction authorization` - ability to update titles
Make title and subtitle customizable so context can be updated.
#### Why is this needed:
image

ivan-zynesis:

/kind feature

Added loading message option into option (pushed to store)
bug fix - not using message pushed from caller

Fixes #496 
Fixes #560

### DIFF
--- a/app/screens/AppNavigator/screens/Settings/SettingsScreen.tsx
+++ b/app/screens/AppNavigator/screens/Settings/SettingsScreen.tsx
@@ -34,12 +34,13 @@ export function SettingsScreen ({ navigation }: Props): JSX.Element {
     }
 
     const auth: Authentication<string[]> = {
-      message: translate('screens/Settings', 'To continue downloading your recovery words, we need you to enter your passcode.'),
       consume: async passphrase => await MnemonicStorage.get(passphrase),
       onAuthenticated: async (words) => {
         navigation.navigate({ name: 'RecoveryWordsScreen', params: { words }, merge: true })
       },
-      onError: e => Logging.error(e)
+      onError: e => Logging.error(e),
+      message: translate('screens/Settings', 'To continue downloading your recovery words, we need you to enter your passcode.'),
+      loading: translate('screens/Settings', 'Decrypting recovering words...')
     }
     dispatch(authentication.actions.prompt(auth))
   }, [walletContext.wallets[0]])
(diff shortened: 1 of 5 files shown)


app/screens/AppNavigator/screens/Settings/SettingsScreen.tsx
```diff
@@ -34,12 +34,13 @@ export function SettingsScreen ({ navigation }: Props): JSX.Element {
...
 
     const auth: Authentication<string[]> = {
-      message: translate('screens/Settings', 'To continue downloading your recovery words, we need you to enter your passcode.'),
       consume: async passphrase => await MnemonicStorage.get(passphrase),
       onAuthenticated: async (words) => {
         navigation.navigate({ name: 'RecoveryWordsScreen', params: { words }, merge: true })
       },
-      onError: e => Logging.error(e)
+      onError: e => Logging.error(e),
+      message: translate('screens/Settings', 'To continue downloading your recovery words, we need you to enter your passcode.'),
```
thedoublejay, line 42: Maybe it would be better if we don't use downloading as users are not downloading a copy. To continue viewing your recovery words...

app/screens/AppNavigator/screens/Settings/SettingsScreen.tsx
```diff
@@ -34,12 +34,13 @@ export function SettingsScreen ({ navigation }: Props): JSX.Element {
...
+      loading: translate('screens/Settings', 'Decrypting recovering words...')
```
thedoublejay, line 43: Don't use "decrypting" as it might alarm security that their wallet will be unsafe. Let's keep it generic like Loading...

app/screens/AppNavigator/screens/Settings/screens/ConfirmPinScreen.tsx
```diff
@@ -36,7 +36,7 @@ export function ConfirmPinScreen ({ route }: Props): JSX.Element {
     }
 
     const copy = { words, network, pin }
-    setSpinnerMessage(translate('screens/PinConfirmation', 'Updating pin code...'))
+    setSpinnerMessage(translate('screens/PinConfirmation', 'Reencrypting wallet...'))
```
thedoublejay, line 39: Same here, let's use updating passcode instead of reencrypting as their wallets didn't lose encryption

thedoublejay (changes requested):

thedoublejay (approved):

PR #630, opened 2021-08-17
`e2e` - updates tests

thedoublejay:

/kind chore

2. Smoke - Change Passcode
3. Smoke - Check if API returns same value in DEX PoolPairs
4. Added tests for API errors
5. Updated Add Liquidity Tests
6. Updated Remove Liquidity Tests

Fixes #


cypress/integration/functional/wallet/balances/balances.spec.ts
```diff
@@ -26,3 +28,21 @@ context('Wallet - Balances', () => {
     cy.getByTestID('address_text').should('exist')
   })
 })
+
+context.only('Wallet - Balances - Failed API', () => {
```
ivan-zynesis, line 32: is the `.only` here on purpose?

cypress/integration/functional/wallet/dex/poolswap.spec.ts
```diff
@@ -87,3 +87,21 @@ context('Wallet - DEX - Pool Swap with balance', () => {
     })
   })
 })
+
+context.only('Wallet - DEX - Pool Swap failed api', () => {
```
ivan-zynesis, line 91: another .only

cypress/integration/functional/wallet/balances/balances.spec.ts
thedoublejay: ha

thedoublejay: my bad haha I'm testing this case only

.gitignore
```diff
@@ -48,3 +48,5 @@ web-build/
 # Cypress
 /cypress/screenshots
 /cypress/videos
+
+/cypress-visual-screenshots/**
```
fuxingloh, line 52: the jpg to check against need to be committed

thedoublejay: @fuxingloh the QR code varies per run. It creates the snapshot first -> exit wallet -> load and check the snapshot

fuxingloh: Oh, I think we should have a fixed entered mnemonic and check against the QR code to detect regression in QR generation.

cypress/integration/functional/wallet/balances/receive.spec.ts
```diff
@@ -25,3 +25,16 @@ context('Wallet - Receive', () => {
...
+context('Wallet - Receive - QR Code - Check', () => {
+  before(function () {
+    cy.createEmptyWallet()
+    cy.getByTestID('balances_list').should('exist')
+    cy.getByTestID('balances_row_0_utxo').click()
+    cy.getByTestID('receive_button').click()
+  })
+
+  it('should match QR code', function () {
+    cy.getByTestID('qr_code_container').compareSnapshot('qr-code-container')
```
fuxingloh, line 38: should check the address also, making sure it's the same address

thedoublejay: this one is handled on another spec

fuxingloh (approved):

PR #632, opened 2021-08-17
Privacy lock

### LINKED ISSUE #434: `biometric` - display unlock screen
Wallet should have an unlock step before opening for privacy purposes
/triage accepted
/area ui-ux

ivan-zynesis:

/kind feature

- refactor LocalAuthContext to be wallet independent
- privacy lock implementation - toggle in Settings, prompt for authentication on app focused

Fixes #434 

#### Additional comments?:
Pending for design and copies confirmation.

Preview:
When no biometric sensor or phone did not enable any
image

Current implementation, directly triggered from setting page. Similar overlay when app focused.
image

Enabled
image


app/screens/PrivacyLock.tsx
```diff
@@ -0,0 +1,22 @@
...
+          const backHandler = BackHandler.addEventListener('hardwareBackPress', () => null)
+          localAuth.privacyLock()
+            .catch(e => BackHandler.exitApp())
+            .finally(() => backHandler.remove())
+        }
+      }
+    })
+  }, [localAuth.isPrivacyLock])
+
+  return null // simplified, not "hiding" UI when authenticating
```
ivan-zynesis, line 21: OS's biometric auth UI will be overlayed on top.

App.tsx
```diff
@@ -33,19 +34,21 @@ export default function App (): JSX.Element | null {
...
-              <StoreProvider>
-                <StatsProvider>
-                  <Main />
-                </StatsProvider>
-              </StoreProvider>
-            </WalletPersistenceProvider>
-          </DeFiScanProvider>
-        </WhaleProvider>
-      </NetworkProvider>
+      <LocalAuthContextProvider>
```
ivan-zynesis, line 37: this is hardware is only dependency and serve as a gate on top on wallet

cypress/integration/functional/wallet/settings/privacyLock.spec.ts
```diff
@@ -0,0 +1,22 @@
+function enablePrivacyLock (): void {
+  localStorage.setItem('Development.Local.PRIVACY_LOCK.enrolled', 'TRUE')
+  cy.wait(500)
+  expect(localStorage.getItem('Development.Local.PRIVACY_LOCK.enrolled')).equal('TRUE')
+}
+
+context('Wallet - Privacy Lock', () => {
```
ivan-zynesis, line 7: Quite minimal tests here, with artificially "enabling" privacy lock in web env which has no biometric hardware.

thedoublejay (changes requested):
1. Doesn't unlock when enabled on existing wallet. Infinite loop
2. Don't lock on focus out. Set a timeout or only on cold start
3. Privacy lock from settings page should be a switch instead of caret. Caret means it needs to go to a page. We just have enable/disable for it
4. If we don't have hardware, is it possible to use phone pattern instead?
5. If there's no hardware, don't display this option

For e2e:
1. We can add a temporary screen on web to test this capability. We need to add these scenarios for e2e

Tests:
1. Initial tests
	- Create wallet 
	- Enable Privacy from Settings
	- Reload page
	- Since wallet is enabled, it's expected to pop up window page
2. On the same context, 
	- Exit on Wallet
	- Create a new Wallet
	- Reload page
	- Privacy lock should be displayed
3. Switch Network
	- If wallets are existing on networks
	- Switch networks
	- Privacy lock should still work
4. On exit wallet
	- Restore the initial wallet
	- Privacy lock should be displayed
5. Hide privacy lock if hardware doesn't exist

BTW, @jtlntno-cake let's not use a OS specific naming (e.g, face id etc.) as it might affect app reviews. Let's just use privacy lock and biometric as generic terms.

image

jtlntno-cake:
> 1. Doesn't unlock when enabled on existing wallet. Infinite loop
> 2. Don't lock on focus out. Set a timeout or only on cold start
> 3. Privacy lock from settings page should be a switch instead of caret. Caret means it needs to go to a page. We just have enable/disable for it
> 4. If we don't have hardware, is it possible to use phone pattern instead?
> 5. If there's no hardware, don't display this option
> 
> For e2e:
> 
> 1. We can add a temporary Alert (we can use WalletAlert) just to test feedback for biometric. We need to add these scenarios for e2e
> 
> Tests:
> 
> 1. Initial tests
>    
>    * Create wallet
>    * Enable Privacy from Settings
>    * Reload page
>    * Since wallet is enabled, it's expected to pop up window page
> 2. On the same context,
>    
>    * Exit on Wallet
>    * Create a new Wallet
>    * Reload page
>    * Privacy lock should be displayed
> 3. Switch Network
>    
>    * If wallets are existing on networks
>    * Switch networks
>    * Privacy lock should still work
> 4. On exit wallet
>    
>    * Restore the initial wallet
>    * Privacy lock should be displayed
> 5. Hide privacy lock if hardware doesn't exist
> 
> BTW, @jtlntno-cake let's not use a OS specific naming (e.g, face id etc.) as it might affect app reviews. Let's just use privacy lock and biometric as generic terms.
> 
> image

1. We should use the toggle-list for the privacy lock, instead of adding another page within the selection.
2. If we detect that the hardware used by user is not compatible with any biometric lock, we can hide this button.
3. @thedoublejay  How about if we use 'Secure with biometric" as the label here

image
image

ivan-zynesis:
> 1. Doesn't unlock when enabled on existing wallet. Infinite loop
Need investigation. I managed to have it working on existing wallet.
Could be build/live refresh issue.

> 2. Don't lock on focus out. Set a timeout or only on cold start
Can try, might have some inconsistency due to app in background, maybe will store the "focus out timestamp" in persistent storage.

> 3. Privacy lock from settings page should be a switch instead of caret. Caret means it needs to go to a page. We just have enable/disable for it
Got it.

> 4. If we don't have hardware, is it possible to use phone pattern instead?
Yes

> 5. If there's no hardware, don't display this option
Since phone pin/pattern works, basically it works for every phone except when the device is not `protected` (swipe to unlock without security)

ivan-zynesis:
@jtlntno-cake 
> If we detect that the hardware used by user is not compatible with any biometric lock, we can hide this button.
@thedoublejay How about if we use 'Secure with biometric" as the label here

It's possible to use phone pin/pattern, do we want to include it?

thedoublejay:
@ivan-zynesis That would be great if it's possible. Need your thoughts also on this @jtlntno-cake  @fuxingloh

ivan-zynesis:
> @ivan-zynesis That would be great if it's possible. Need your thoughts also on this @jtlntno-cake @fuxingloh

That's my understanding based on what I learn from documentation but not tested on any phone without any biometric device.

cypress/integration/functional/wallet/settings/privacyLock_with_stub.spec.ts
```diff
@@ -0,0 +1,39 @@
+// import * as LocalAuthentication from 'expo-local-authentication'
+
+context.skip('Wallet - Privacy Lock', () => {
+  before(function () {
+    /**
+     * FIXME:
+     * stub LocalAuthentication require recompile
+     * error: `require appropriate loader` when local authentication import direct from react-native
+     */
+    // cy.stub(LocalAuthentication, 'hasHardwareAsync').returns(true)
```
ivan-zynesis, line 10: @thedoublejay this is what I mentioned to you earlier.

When import + stub happened within the cypress main, it will cause some changes to the compiling process. Few errors I bumped into and haven't managed to solve them. Possibly easy solve with the correct cypress compiler option, but will takes me time learn through the cypress doc and some trial and error.

ivan-zynesis: image

app/screens/PrivacyLock.tsx
```diff
@@ -1,34 +1,68 @@
...
-          .then()
+  /**
+   * a dynamic event listener handling
+   * array of max length 2
+   * [0] get auto de-registered and spliced from the array when [1] pushed in
+   * as handler itself is callback with dependency, unable to bind a single instance or any singleton pattern
+   */
+  const prevHandlerRef = useRef<Array<(s: AppStateStatus) => void>>([])
+
+  const appStateHandler = useCallback((nextState: AppStateStatus) => {
```
ivan-zynesis, line 18: Split out to 3 different dependency hooks

1. handler itself - depend on privacy enabled status only
2. add app state listener - when handler (1) changed
3. remove previous listener - when a new handler registered (2)

some manual testing will be needed as OSs can behave differently.
- cold start
- app go background then come back (dev build timeout is 3 sec, preview will be 1 minute though, extra difficulty to do test)
- enable privacy lock in this launch (minimize, come back, require authentication)
- disable privacy lock in this launch (minimize, come back, do not require authentication)

app/screens/PrivacyLock.tsx
```diff
@@ -61,6 +61,9 @@ export function PrivacyLock (): JSX.Element | null {
 function authenticateOrExit (localAuth: LocalAuthContext): void {
   const backHandler = BackHandler.addEventListener('hardwareBackPress', () => null)
   localAuth.privacyLock()
-    .catch(e => BackHandler.exitApp())
+    .catch(async () => {
+      await AppLastActiveTimestamp.forceRequireReauthenticate()
```
ivan-zynesis, line 65: The key change is here, when user failed to authenticate. Force the timeout back to older timestamp, thus user cannot skip authentication if resume app within 1minute.

app/api/app_last_active.ts
```diff
@@ -1,18 +1,28 @@
...
+  return await StorageAPI.setItem(KEYS.timestamp, `${Date.now()}`)
+}
+
+async function removeForceAuth (): Promise<void> {
+  return await StorageAPI.removeItem(KEYS.force)
 }
 
 async function forceRequireReauthenticate (): Promise<void> {
-  return await StorageAPI.setItem(KEY, `${Date.now() - getTimeoutPeriod()}`)
+  return await StorageAPI.setItem(KEYS.force, 'TRUE')
```
ivan-zynesis, line 18: Need a higher priority persistent flag. The timestamp tends to get overwrite as the component lifecycle is not very controllable. (The focus out callback reset my "force" timestamp when I programmatically exit app)

app/screens/PrivacyLock.tsx
ivan-zynesis: more test case (timeout mechanics)
- authenticate success, minimize, comeback after 1 minute: re-auth required
- authenticate success, minimize, come back within 1 minute: no re-auth required
- authenticate fail, minimize, come back within 1 minute: auth required

app/contexts/AppStateContext.tsx
```diff
@@ -0,0 +1,85 @@
...
+
+export type SimplifiedAppStateStatus = 'active' | 'background'
+export type AppStateListener = (status: SimplifiedAppStateStatus) => void
+
+export interface AppStateContext {
+  addListener: (listener: AppStateListener) => number
+  removeListener: (id: number) => void
+}
+
+const appStateContext = createContext<AppStateContext>(undefined as any)
```
fuxingloh, line 12:
```suggestion
const AppStateContext = createContext<AppStateContext>(undefined as any)
```

app/contexts/AppStateContext.tsx
```diff
@@ -0,0 +1,85 @@
...
+  // DO NOT use `useState`, these are meant for appstate monitoring
+  // re-render side effect is undesireable
+  const listenerRef = useRef<{ [key: number]: AppStateListener } | null>(null)
+  const prevState = useRef<AppStateStatus>('active') // first mounted
+
+  // all useCallback / useEffect in this context provider MUST have ZERO dependencies, to remain predictable
+  const emit = useCallback((listeners: AppStateListener[], nextState: SimplifiedAppStateStatus) => {
+    prevState.current = nextState // simplify iOS lifecyle, skipping `inactive` as transition state
+    listeners.forEach(l => l(nextState))
+  }, [])
```
fuxingloh, line 28: No point using `useCallback` if no dependencies were passed in, maybe just use a function? 

> Easier to understand, I was trying to understand why we needed a callback.

app/contexts/AppStateContext.tsx
```diff
@@ -0,0 +1,85 @@
...
+
+  const appStateHandler = useCallback((nextState: AppStateStatus) => {
```
fuxingloh, line 30: Same for this

app/contexts/AppStateContext.tsx
```diff
@@ -0,0 +1,85 @@
...
+  }
+
+  return (
+    <appStateContext.Provider value={context}>
+      {props.children}
+    </appStateContext.Provider>
+  )
+}
+
+function getNewSerialNumber (existing: number[]): number {
```
fuxingloh, line 79: use a random number?

app/contexts/AppStateContext.tsx
```diff
@@ -0,0 +1,85 @@
...
+const appStateContext = createContext<AppStateContext>(undefined as any)
+
+export function useAppStateContext (): AppStateContext {
+  return useContext(appStateContext)
+}
+
+export function AppStateContextProvider (props: React.PropsWithChildren<any>): JSX.Element | null {
+  // DO NOT use `useState`, these are meant for appstate monitoring
+  // re-render side effect is undesireable
+  const listenerRef = useRef<{ [key: number]: AppStateListener } | null>(null)
```
fuxingloh, line 21:
```suggestion
  const listenerRef = useRef<Record<number, AppStateListener>>({})
```

keep this initialized? reducing the need for `if (listenerRef !=)`

app/contexts/AppStateContext.tsx
```diff
@@ -0,0 +1,85 @@
...
+      if (prevState.current === 'background') {
+        emit(listeners, 'active')
+      }
+    }
+  }, [])
+
+  useEffect(() => {
+    const cb = (ns: AppStateStatus): void => {
+      appStateHandler(ns)
+    }
```
fuxingloh, line 46:
```suggestion
    function resolveState(): string {
      ...
    }

    function onChange(status: AppStateStatus): void {
      prevState.current = status
      const state = resolvedState(...)
      
      for (listener ... ) {
        listener(state)
      }
    }
```

Should refactor and simplify this code, unnecessarily complex. From where I see, you just need to:

1. store prev state
2. resolve new state
3. for each emit out the newly resolved state

app/contexts/LocalAuthContext.tsx
```diff
@@ -0,0 +1,110 @@
+import { Logging } from '@api'
+import * as LocalAuthentication from 'expo-local-authentication'
+import { AuthenticationType, LocalAuthenticationOptions, SecurityLevel } from 'expo-local-authentication'
+import React, { createContext, useCallback, useContext, useEffect, useState } from 'react'
+import { PrivacyLockPersistence } from '../api/wallet/privacy_lock'
+
+export interface LocalAuthContext {
```
fuxingloh, line 7:
```suggestion
export interface LocalAuthContextI {
```

app/contexts/LocalAuthContext.tsx
```diff
@@ -0,0 +1,110 @@
...
+
+  // API
+  isPrivacyLock: boolean
+  privacyLock: (options?: LocalAuthenticationOptions) => Promise<void>
+  enablePrivacyLock: (options?: LocalAuthenticationOptions) => Promise<void>
+  disablePrivacyLock: (options?: LocalAuthenticationOptions) => Promise<void>
+  togglePrivacyLock: (options?: LocalAuthenticationOptions) => Promise<void>
+}
+
+const localAuthContext = createContext<LocalAuthContext>(undefined as any)
```
fuxingloh, line 23:
```suggestion
const LocalAuthContext = createContext<LocalAuthContext>(undefined as any)
```

app/contexts/LocalAuthContext.tsx
```diff
@@ -0,0 +1,110 @@
...
+      })
+  }, [])
+
+  // native expo-local-authenticate
+  const _authenticate = useCallback(async (options?: LocalAuthenticationOptions) => {
+    const result = await LocalAuthentication.authenticateAsync(options)
+    if (!result.success) {
+      throw new Error(result.error)
+    }
+  }, [])
```
fuxingloh, line 61: don't use callback if there isn't any deps

app/contexts/LocalAuthContext.tsx
```diff
@@ -0,0 +1,110 @@
...
+    disablePrivacyLock: async (options) => await setPrivacyLock(false, options),
+    togglePrivacyLock: async (options) => {
+      if (isPrivacyLock === true) {
+        return await context.disablePrivacyLock(options)
+      }
+      return await context.enablePrivacyLock(options)
+    }
+  }
+
+  return (
```
fuxingloh, line 105: delay returning of children until LocalAuthContext has been properly loaded.

app/contexts/LocalAuthContext.tsx
```diff
@@ -0,0 +1,110 @@
...
+  hardwareSecurityLevel: SecurityLevel
+  supportedTypes: AuthenticationType[]
+  isDeviceProtected: boolean
+
+  // API
+  isPrivacyLock: boolean
+  privacyLock: (options?: LocalAuthenticationOptions) => Promise<void>
+  enablePrivacyLock: (options?: LocalAuthenticationOptions) => Promise<void>
+  disablePrivacyLock: (options?: LocalAuthenticationOptions) => Promise<void>
+  togglePrivacyLock: (options?: LocalAuthenticationOptions) => Promise<void>
```
fuxingloh, line 20: I think you should reduce the interface, too much IMO.

1. support privacy lock?
2. type of supported privacy lock (choose 1)
3. isEnabled?
4. setEnabled(boolean)

fuxingloh: And should rename to `PrivacyLockContext`.

app/contexts/AppStateContext.tsx
ivan-zynesis: erm, not sure if I misunderstand some react lifecycle. `useCallback` with no dependency so that it will not get reinstantiate on re-render.

app/contexts/LocalAuthContext.tsx
ivan-zynesis: Reduced API provided, but left those hardware status getter there, will be needed if UI trying to explain why privacy lock option is disabled.

app/contexts/LocalAuthContext.tsx
```diff
@@ -0,0 +1,109 @@
...
+import React, { createContext, useCallback, useContext, useEffect, useState } from 'react'
+import { PrivacyLockPersistence } from '../api/wallet/privacy_lock'
+
+export interface PrivacyLockContextI {
+  // user's hardware condition, external
+  fetchHardwareStatus: () => void // not likely needed, in case user change device's security setting
+  hasHardware: boolean
+  hardwareSecurityLevel: SecurityLevel
+  supportedTypes: AuthenticationType[]
+  isDeviceProtected: boolean
```
ivan-zynesis, line 13: Reduced some of the API function, leaving hardware status getter here, they are required if UI trying to explain why privacy lock option is disabled for requirement not met situation.

app/api/app_last_active.ts
```diff
@@ -0,0 +1,50 @@
...
+
+/**
+ * Failed passcode input counter persistence layer
+ */
+export const AppLastActiveTimestamp = {
+  set,
+  forceRequireReauthenticate,
+  removeForceAuth,
+  shouldReauthenticate
+}
```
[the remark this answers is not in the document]
ivan-zynesis: Trying with it now, remove the code after tested. Not certain on the memory management on different OS, persistent storage was the one we can confirm the behavior.

app/api/app_last_active.ts
```diff
@@ -0,0 +1,50 @@
+import { EnvironmentName, getEnvironment } from '@environment'
```
fuxingloh, line 1: do we still need this file?

ivan-zynesis: Nope, will delete all the `persistent layer` for `last active` after some manual test across platforms are done. Currently the usage are commented out.

izzycsy:
> * No wallets loaded - No Prompt on Close and Timeout

I’ve amended some tests to according to several UX scenarios, and added more test scenarios as well.

fuxingloh (changes requested):
On cancel should hide balances, the current state management in PrivacyLock only seem to hide when Authentication Screen is up.


https://user-images.githubusercontent.com/4266087/133629337-805bb7aa-bf61-4b18-b5e5-707af9c8f472.mp4

app/screens/AppNavigator/screens/Settings/SettingsScreen.tsx
```diff
@@ -207,6 +222,28 @@ function RowExitWalletItem (): JSX.Element {
...
+  onToggle
+}: { disabled?: boolean, value: boolean, onToggle: (newValue: boolean) => void }): JSX.Element {
+  return (
+    <ThemedView
+      light={tailwind('bg-white border-b border-gray-200')}
+      dark={tailwind('bg-gray-800 border-b border-gray-700')}
+      style={tailwind('flex p-4 pr-2 flex-row items-center justify-between')}
+    >
+      <ThemedText testID='text_privacy_lock' style={tailwind('font-medium')}>
+        {translate('screens/Settings', 'Privacy Lock')}
```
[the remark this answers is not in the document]
thedoublejay: @fuxingloh we have discussed it before, but we decided not to put any OS specific features (e.g, FaceID/TouchID for Apple etc.) since it might affect store approval.

thedoublejay:
@fuxingloh I cannot reproduce the overlay issue on iOS and Android. But I've added a condition on the overlay to show on cancelled actions https://github.com/DeFiCh/wallet/pull/632/commits/69b47f998e0c376ac814c359bec763ddc49907a7

app/screens/AppNavigator/screens/Settings/SettingsScreen.tsx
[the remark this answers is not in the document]
fuxingloh: https://developer.apple.com/design/human-interface-guidelines/ios/user-interaction/authentication/

> Always identify the authentication method. A button for signing in to your app using Face ID, for example, should be titled "Sign In with Face ID" rather than "Sign In."

Their guidelines require us to mention it for Apple.

fuxingloh:
We should add an alert stating how the privacy lock work, e.g. how many seconds etc

fuxingloh:
> @fuxingloh I cannot reproduce the overlay issue on iOS and Android. But I've added a condition on the overlay to show on cancelled actions [69b47f9](https://github.com/DeFiCh/wallet/commit/69b47f998e0c376ac814c359bec763ddc49907a7)

I double-checked with this commit it's resolved.

fuxingloh (approved):

thedoublejay (approved):

PR #633, opened 2021-08-17
`remove liquidity` - add input text for percentage 

### LINKED ISSUE #595: `remove liquidity` - add input text for percentage

kyleleow:

/kind feature

- To add another way to input percentage instead of only slider in Remove Liquidity Screen

Fixes #595

#### Additional comments?:


app/screens/AppNavigator/screens/Dex/DexRemoveLiquidity.tsx
```diff
@@ -70,20 +73,47 @@ export function RemoveLiquidityScreen (props: Props): JSX.Element {
       .catch(Logging.error)
   }, [])
 
+  useEffect(() => {
+    setValidity(
+      Number(percentage) > 0 &&
```
thedoublejay, line 78: Can we use BigNumber for this?

app/screens/AppNavigator/screens/Dex/DexRemoveLiquidity.tsx
```diff
@@ -70,20 +73,47 @@ export function RemoveLiquidityScreen (props: Props): JSX.Element {
...
-            style={tailwind('flex-1 font-semibold')}
+            style={tailwind('w-2/4 font-semibold flex-1')}
           >{translate('screens/RemoveLiquidity', 'Amount to remove')}
           </Text>
-          <Text testID='text_slider_percentage' style={tailwind('text-right')}>{percentage}%</Text>
+          <NumberTextInput
+            testID='text_input_percentage'
+            style={[
+              tailwind('text-right w-2/4 p-0 mr-0.5'),
+              Platform.OS === 'ios' && tailwind('-mt-0.5'),
```
thedoublejay, line 96: store this on a boolean instead of inline, so it's easier for context

thedoublejay (changes requested):

app/screens/AppNavigator/screens/Dex/DexRemoveLiquidity.tsx
kyleleow: you mean this `Platform.OS === 'ios'` right? Should i refactor to a constant? So it will be available project-wide

thedoublejay: no no, just put it as a const above the return. `isIOS` then use that boolean here instead of inlining the expression

kyleleow: yes but i noticed the same inline code is used in HeaderTitle too, should we update that?

thedoublejay (approved):

thedoublejay (approved):

PR #657, opened 2021-08-18
updated welcome page

thedoublejay:

/kind feature

Fixes #344 

<div>

</div>


fuxingloh (approved):
maybe we should precache the image, due to the load time

PR #665, opened 2021-08-19
`Store/transactionNotification` - for await transaction confirmation

### LINKED ISSUE #640: `Ocean Interface`: wait for confirmation should listen for block activity

Wait for confirmation should check for transaction immediately after a block update via `StatsProvider`. As new blocks equal new transactions getting confirmed.

/area ui-ux
/triage accepted

ivan-zynesis:

/kind feature

Centralized transaction on chain status checker via redux store.
Triggered by StatsProvider, on each new block found.

Fixes #640


thedoublejay (changes requested):
With this implementation, did we lose the ability to have a timeout for confirmation? Confirmation on Mainnet takes time. We should still have a timeout of 5 minutes. Their TX will be broadcasted but not confirmed. All our operations are disabled if there's a pending transaction. So without a timeout, in case there will be some unexpected issue, they won't be able to resume operations.

ivan-zynesis:
> With this implementation, did we lose the ability to have a timeout for confirmation? Confirmation on Mainnet takes time. We should still have a timeout of 5 minutes. Their TX will be broadcasted but not confirmed. All our operations are disabled if there's a pending transaction. So without a timeout, in case there will be some unexpected issue, they won't be able to resume operations.

Nope, I would expect it continue working. But I have tested only for playground, it works. 
submit tx -> subscribe notification -> block update (stat provider) -> transaction notification -> dismiss ocean drawer

But leaving in app in background for 30 mins may fail since OS may have take them out, but I think it might be the same with or without this PR change.

PR #666, opened 2021-08-19
`dex` - add skeleton loader

### LINKED ISSUE #609: `skeleton loader` - initial implementation
Do an initial implementation of skeleton loader on one page. So we can test the look and feel first before fully implementing everywhere. Start with DEX Screen, Add some skeleton loader on pool pairs while it's still fetching
#### Why is this needed:

Something like this - https://www.npmjs.com/package/react-native-skeleton-content
or this one I think is better - https://www.npmjs.com/package/react-content-loader

kyleleow:

/kind feature

- Add skeleton loader to inform user that the content is loading if the API is taking time to return

Fixes #609

#### Additional comments?:
- Tested on web and pixel 3a (with network throttling) and iPhone 12
- Requires network throttling to test on playground network


thedoublejay (changes requested):
Good start. Some comments:
1. Create a separate skeleton component on the Screen directory. Don't add it on the file itself.
2. On Android, it's not aligned to the right side
Screenshot_20210820-074739_Expo Go
3. Blockers are too rounded. Our overall designs are using a slight round at the corners. Maybe it would be better for us to use plain squares instead of rounded corners. Icons should be rounded but rest can be square
4. Add 3 skeleton items on DEX page since it's a list
5. Need to add e2e case for this. What you can do is to add it on dex spec. Add a new content and make it as skeleton loader check. Add an intercept of the API call and add a delay before responding. Ideally, it should display the content. After the timeout, it should then load the normal display.

kyleleow:
@thedoublejay this is after removing the rounded corners
image

app/screens/AppNavigator/screens/Dex/DexScreen.tsx
```diff
@@ -62,29 +72,38 @@ export function DexScreen (): JSX.Element {
...
+              </View>
+            )
+
+          case SectionKey.AvailablePoolPair:
+            return (
+              <>
+                <SectionTitle text={translate('screens/DexScreen', section.key)} testID={section.key} />
+                {isEmpty(section.data) && (
+                  <>
+                    <DexSkeletonLoader />
```
thedoublejay, line 99: Add a key here

kyleleow: updated

app/screens/AppNavigator/screens/Dex/DexScreen.tsx
```diff
@@ -209,3 +222,14 @@ function PoolPairAPR (props: { symbol: string, apr: number, row: string }): JSX.
     </View>
   )
 }
+
+function DefaultSkeletonLoader (): JSX.Element {
+  const skeletonRow = 3
+  return (
+    <>
+      {Array.from(Array(skeletonRow), (_v, i) => i + 1).map((_x, i) => (
```
thedoublejay, line 230: this is too complex, haha just make it 

const skeletonRow = [1,2,3]

skeletonRow.map((x) => <skeletonrow key={x} />)

thedoublejay (changes requested):

app/screens/AppNavigator/screens/Dex/DexScreen.tsx
kyleleow: 😅 updated

thedoublejay (approved):

PR #678, opened 2021-08-21
dark mode

### LINKED ISSUE #565: `dark mode` - initial implementation

- [ ] Adjust based on System preferences
- [ ] Change to Light
- [ ] Change to Dark
- [ ] Dynamic splash screen (?)

#### Why is this needed:
@jtlntno-cake might need your input on this

thedoublejay:

/kind feature

1. Ability to select dark mode
2. Checks default theme
3. Saves to Storage from Settings Page

Fixes #565


app/contexts/ThemeProvider.tsx
```diff
@@ -0,0 +1,97 @@
...
+  isThemeLoaded: boolean
+}
+
+export function useTheme (): ThemeLoader {
+  const colorScheme = useColorScheme()
+  const [isThemeLoaded, setIsThemeLoaded] = useState<boolean>(false)
+  const [theme, setTheme] = useState<NonNullable<ColorSchemeName>>('light')
+
+  useEffect(() => {
+    ThemePersistence.get().then((t) => {
```
fuxingloh, line 17: Maybe better if use just use system provided theme instead of having any overriding mechanism.

Similar to Language, always use system default.

app/contexts/ThemeProvider.tsx
```diff
@@ -0,0 +1,97 @@
...
+  'body-text': 'text-white text-opacity-90',
+  'subtitle-text': 'text-white text-opacity-70',
+  'text-primary': 'text-darkprimary',
+  'light-text': 'text-gray-300'
+}
+
+interface Theme {
+  theme: NonNullable<ColorSchemeName>
+  setTheme: (theme: NonNullable<ColorSchemeName>) => void
+  getThemeClass: (type: string) => string
```
fuxingloh, line 59:
```suggestion
  tailwind: function
```

allowing: 
```ts
const { tailwind } = useTheme()
```

We should do this instead, let the `Theme` Context automatically convert without having to use `getThemeClass`. 

By introducing `getThemeClass` you introduce the CSS Zen Garden concept which is an anti pattern of utility first css - tailwind. This sort of approach is hard to manage and scale leading up to a concentration of semantic to maintain. Who maintain this semantics, how does it evolve, does it break another screen? Leading up to the problems which leds to the eventual creation of utility first css.

@aentan shared this a few months back:
> https://adamwathan.me/css-utility-classes-and-separation-of-concerns/
> IMO this is the only sure way to really get UI consistency and polish. In my 18 years writing FE code, I’ve done it all. CSS Zen Garden semantic CSS, BEM, etc. I’ve arrived at the same conclusion as the above author.

I share the same opinion as @aentan

fuxingloh (changes requested):

app/contexts/ThemeProvider.tsx
thedoublejay: It always use system default. The overwriting only happens when user explicitly update it from settings page. So when they re-open the app, it should keep the change that they added.

Same for Language, it should be system default but we should still allow users to select and replace the default loaded if they want.

thedoublejay:
@jtlntno-cake test

PR #691, opened 2021-08-24
`transaction` - add skeleton loader

### LINKED ISSUE #690: `transaction` - add skeleton loader
- Display skeleton loader when:
  - first load
  - pull to refresh
- Background update should not trigger skeleton loader

#### Why is this needed:
- To inform user that the data is loading

Transaction Screen

kyleleow:

/kind feature

- Add skeleton loader to inform user that data is loading
- Fix "ghost pulling" that was caused by background update triggerring refresh action

Fixes #690

#### Additional comments?:


app/store/transaction.ts
```diff
@@ -0,0 +1,42 @@
+import { createSlice, PayloadAction } from '@reduxjs/toolkit'
+import { VMTransaction } from '../screens/AppNavigator/screens/Transactions/screens/stateProcessor'
+
+type LoadingState = 'idle' | 'loading' | 'loadingMore' | 'success' | 'background' | 'error'
+
+export interface TransactionState {
```
thedoublejay, line 6: Why did you add it on the global state? It's not used anywhere else except Transaction Page

thedoublejay (changes requested):

app/store/transaction.ts
kyleleow: My intention is to move the side effects that is not related to the rendering logic outside from the component. 

But I couldn't achieve what I wanted to do when I tried to make async call using redux's createAsyncThunk(). Hence I settled with this solution.

Will there be any scenario or feature in the future that we might need to access transaction data anywhere else? If there is none then I'll move the state out from redux.

thedoublejay: Nope, we will only use it on this page. Yes please move it out of the global store

kyleleow: Updated

cypress/integration/functional/wallet/transaction/list.spec.ts
```diff
@@ -1,10 +1,29 @@
 context('Wallet - Transaction - List', () => {
   describe('wallet has 2 transactions in history', function () {
     before(() => {
+      cy.visit('/')
```
thedoublejay, line 4: @kyleleow this one is already done inside createEmptyWallet

thedoublejay (changes requested):

cypress/integration/functional/wallet/transaction/list.spec.ts
kyleleow: updated

thedoublejay (approved):

PR #692, opened 2021-08-24
use different DfTx when max DFI sent

### LINKED ISSUE #683: `send` - cannot send max DFI
#### What happened:
Cannot send max amount of UTXO DFI. Need to check if we're using https://github.com/DeFiCh/jellyfish/blob/0b61eaee5afba588b819c923cda762440f3ced5c/packages/jellyfish-transaction-builder/src/txn/txn_builder_utxo.ts#L17

ivan-zynesis:

/kind fix

Fixes #683


thedoublejay (changes requested):
@ivan-zynesis do we use the same method for convert? Maybe we can use it also

thedoublejay (approved):

PR #701, opened 2021-08-25
Updated Balance, Token detail and Send token screen to differentiate wrapped and native tokens

### LINKED ISSUE #555: token symbols - append `d` for wrapped tokens and update logos

- [x] For all wrapped tokens that we have, add a prefix 'd' to differentiate wrapped and native tokens for atomic swap
- [x] Update icons for wrapped tokens (Lighter) @jtlntno-cake  to provide new tokens

#### Why is this needed:
image

fullstackninja864:

/kind fix

For all wrapped tokens that we have, added a prefix 'd' to differentiate wrapped and native tokens for  Balance, Token detail and Send token screen

Fixes #555


fullstackninja864:
Closing The PR as discussed https://github.com/DeFiCh/wallet/issues/555#issuecomment-905307527

PR #727, opened 2021-08-27
`convert` - auto deduct 0.1 DFI UTXO when converting max amount

### LINKED ISSUE #715: `convert` - on max convert UTXO -> Token, auto deduct 0.1 DFI
On max convert of UTXO -> Token, auto deduct 0.1 DFI to cover transaction fees. This will cause some e2e cases to be updated since we are expecting it to throw an error.

kyleleow:

/kind feature

- To auto deduct 0.1 DFI UTXO if user try to convert all UTXO to Token

Fixes #715 

#### Additional comments?:
- Added condition to block continue button if balance after convert is less than 0.1
- Added condition to set UTXO input amount 0 if user press on 50% or MAX button when UTXO balance is equal or less than 0.1
- Tested on iPhone 12 and pixel 3a emulator

### DIFF
--- a/app/screens/AppNavigator/screens/Balances/screens/ConvertScreen.tsx
+++ b/app/screens/AppNavigator/screens/Balances/screens/ConvertScreen.tsx
@@ -49,8 +49,30 @@ export function ConvertScreen (props: Props): JSX.Element {
   const [targetToken, setTargetToken] = useState<ConversionIO>()
   const [convAmount, setConvAmount] = useState<string>('0')
   const [fee, setFee] = useState<BigNumber>(new BigNumber(0.0001))
-
   const [amount, setAmount] = useState<string>('')
+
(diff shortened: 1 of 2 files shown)


thedoublejay:
@kyleleow I think you might have added extra validations. You can reduce the UTXO balance directly on `getDFIBalances`. Subtracting 0.1 will already validate forms and you don't need to add extra validation. Currently, our form has a validation for gte balance, positive number and not zero.

thedoublejay (changes requested):
@kyleleow I think you might have added extra validations. You can reduce the UTXO balance directly on getDFIBalances. Subtracting 0.1 will already validate forms and you don't need to add extra validation. Currently, our form has a validation for gte balance, positive number and not zero.

thedoublejay (approved):

PR #747, opened 2021-08-31
consistent eslint styling

thedoublejay:

/kind chore

- Added rules for react, react-native and react-hooks
- added curly rule - to ensure curly if statements
- brace style - to avoid inline braces
- keyword spacing - consistent spacing after keywords

Fixes #


.eslintrc
```diff
@@ -2,6 +2,11 @@
...
     "project": "./tsconfig.json"
   },
+  "plugins": [
+    "react",
+    "react-native",
+    "react-hooks"
+  ],
   "extends": [
     "standard-with-typescript",
     "standard-jsx"
```
fuxingloh, line 12: need to extend the plugins you added?

- plugin:react-native/all
- plugin:react-hooks/all

thedoublejay: let me check

fuxingloh (approved):

Issue #689, opened 2021-08-24
`poolswap` - DFI - BTC disabled on some cases


thedoublejay:
#### What happened:
When doing poolswap, continue button is disabled and user is not able to proceed. There's a workaround when they restart the app. Needs to be investigated.

thedoublejay:
Very hard to reproduce. Tested both on iOS and Android. Need more details for this. Daniel reported it's very intermittent and it only happen on initial use of app.

Issue #728, opened 2021-08-27
cannot swap DFI to BTC from bigger amount of 1 DFI


BastianReed:
i want to swap my DFI Token into BTC. 
I have 68 DFI Token and 11 DFI UTXO.
If i use more than 1 DFI to swap, the "Continue" Button is displayed grey and not confirmable.

I use the Android Defichain Wallet.

thedoublejay:
Hey @BastianReed, we have an issue regarding this. https://github.com/DeFiCh/wallet/issues/689. It's currently hard to reproduce on our end. Can you give us more info or step by step on how you have encountered this? Also, as a workaround, you may try to restart the app and try again. Some users reporting this only on initial start of app.

thedoublejay:
@BastianReed, also if you can provide us what app version, phone version and model you have so we can check. thanks

thedoublejay:
Hey @BastianReed, this has been fixed with the latest app update `v0.10.5`. Please update your app via App/Google Store.

PR #753, opened 2021-08-31
updated useCallback dependencies

thedoublejay:

/kind refactor

1. Updated useCallback dependencies. All variables which is used inside useCallback should be added on the dependencies. It might be the rootcause of issue #728 
2. Removed indent as it causes conflicts with editors, tslint and eslint. Will check on another PR
3. Remove unnecessary warn for cypress. Waiting is necessary especially for block wait

Fixes https://github.com/DeFiCh/wallet/issues/689


fuxingloh (approved):
One thing to think about is how we manage states across 'forms'.

For `<forms/>`, I don't think we should live read from the store and refresh the form when we detect changes. This changes jitter refresh when the values changes. 

For example: `50% of 1.0 = 0.5` and `50% of 1.1 != 0.5` after update.
While when we enter into `<form/>` state we should not read the value actively from store. We should cache them instead for that transaction regardless of state change.

`1.0` to stay `1.0` even when `0.1 DFI` came in during the process.

Below should be static for the lifecycle of the action.
```ts
  }, [JSON.stringify(tokens), poolpair])
```

thedoublejay:
@fuxingloh  I need to check what will be the changes in case we do it. I remember adding it before some reason. Need to revisit. I'll create a ticket for it for now

PR #767, opened 2021-09-01
`network` - add screen to change network outside of wallet context

### LINKED ISSUE #566: `network` - select outside of wallet context
Currently, when user is out of wallet context (e.g, Onboarding, restore flow). User doesn't have a way to select networks. He needs to create a wallet first and is currently a funnel.

- [ ] Add on Create Wallet Guidelines (First Screen only)
- [ ] Restore page

#### Why is this needed:

kyleleow:

/kind feature

- To provide an option to user to switch network during create wallet and restore wallet flow

Fixes #566

#### Additional comments?:
Replace bottom modal used in mock UI with a screen due to technical difficulty

#### Developer Checklist:

- [x] Unit tests*
- [x] Added e2e tests*
- [x] Added translations*


thedoublejay:
@kyleleow we need to have a different alert message when you're out of wallet context since you're already on the onboarding screen.

kyleleow:
@thedoublejay For the new message, can I go with: "You are about to switch to {{network}}. Do you want to proceed?"

thedoublejay:
Ok @kyleleow, you can ask @izzycsy for the german translations also

kyleleow:
Added german translation from Bella

thedoublejay (changes requested):
1. Error on message when switching both on outside of wallet context and from settings page. Network selecting is wrong. Probably a good idea to add an e2e on this.

thedoublejay:
Also, one more minor change. Can you replace the icon only on Guidelines page and Restore page? Just to signify that the network can be selected. You can check with @jtlntno-cake on the icon class name.
Screenshot 2021-09-02 at 4 57 34 PM

app/components/HeaderTitle.tsx
```diff
@@ -1,10 +1,35 @@
...
+      <ThemedText
+        dark={tailwind('text-white text-opacity-90')}
+        light={tailwind('text-black')}
+        style={tailwind('font-semibold leading-5')}
+        testID={testID}
+      >
+        {text}
+      </ThemedText>
+
+      {subHeadingType === 'None' && <></>}
```
ivan-zynesis, line 27: This is redundant right? undefined should cover it.

app/screens/WalletNavigator/screens/CreateWallet/OnboardingNetworkSelectScreen.tsx
```diff
@@ -0,0 +1,30 @@
...
+  return (
+    <View testID='onboarding_network_selection_screen'>
+      <SectionTitle
+        testID='onboarding_network_selection_screen_title'
+        text={translate('screens/OnboardingNetworkSelectScreen', 'SELECT NETWORK')}
+      />
+
+      {
+        networks.map((network, index) => (
+          <RowNetworkItem
```
ivan-zynesis, line 20: I actually prefer just make another simple component here.

1. This is not part of Settings screen anyway
2. Wish this page page auto dismiss after selected a network (my personal preference on the UX)
3. The original row component will has fixed alert message and tidier

kyleleow: After consulting Jonathan, he mentioned that the network selection UI will have changes again in the future and both screen (onboarding and setting) should always be the same.

Can I still go with the current solution which is to share the same `RowNetworkItem` component, but I'll move it to `app/components` ?

thedoublejay (approved):

PR #786, opened 2021-09-03
`settings` - add language selection screen

### LINKED ISSUE #772: `language selector` - settings page
Currently, we use the language of the phone to update translations. Users are asking if we can add a way to change the language in app. Use the same logic as theme.

- [ ] Load the language based on selected setting on local storage
- [ ] If there's none, load from app settings
- [ ] Add select language screen from Settings Page
- [ ] UI should be like network selection. Display the selected language, on click it should be able to change the languages.
- [ ] On restart of app, we should load the correct language
- [ ] Display German/English for now

kyleleow:

/kind feature

- To provide an option for user to change language within app

Fixes #772

#### Developer Checklist:

- [x] Unit tests*
- [x] Added e2e tests*
- [x] Added translations*


app/screens/AppNavigator/screens/Settings/components/RowLanguageItem.tsx
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  return (
+    <ThemedTouchableOpacity
+      onPress={onPress}
+      style={tailwind('flex flex-row p-4 pr-2 items-center justify-between')}
+      testID={`button_language_${props.language}`}
+    >
+      <ThemedText style={tailwind('font-medium')}>
+        {props.language}
```
thedoublejay, line 55: Language should be translated also. E.g, English is `Englisch` in German.

app/translations/index.ts
```diff
@@ -82,3 +83,25 @@ function deepEncode (obj: any): any {
 export function encodeScope (text: string): string {
   return Buffer.from(text).toString('base64')
 }
+
+export function getLocaleByLanguageName (languageName: EnvironmentLanguage): string {
+  switch (languageName) {
```
thedoublejay, line 88: Would it make sense to use the language key instead of language name as value of the list? Some countries will have same locale key

app/screens/AppNavigator/screens/Settings/components/RowLanguageItem.tsx
kyleleow: @thedoublejay I was referring to the how iOS and FB handle the language selection screen, the main option is displayed in the target language, then the description is using the current chosen language.

Do you want us to follow them as well?

fb_language
ios_language

thedoublejay: Oh ok got it. You can check this with @jtlntno-cake. I'll let him decide which one should we use.

Just as a reference, on the desktop app we are translating every labels

jtlntno-cake: Yeah I like the suggestion. I incorporated that into the screen then.
Here's the [proto](https://www.figma.com/proto/SDIpZciuxJlUf4xf8yheiY/Light-Wallet?node-id=287%3A11410&viewport=241%2C48%2C0.57&scaling=scale-down&starting-point-node-id=841%3A28929&show-proto-sidebar=1)

app/environment.ts
```diff
@@ -16,10 +16,16 @@ export enum EnvironmentName {
...
+export enum EnvironmentLanguage {
+  English = 'English',
+  German = 'Deutsch'
+}
+
 interface Environment {
   name: EnvironmentName
   debug: boolean
   networks: EnvironmentNetwork[]
+  languages: EnvironmentLanguage[]
```
thedoublejay, line 28: Languages shouldn't be an environment level setting. Just use app level setting

app/screens/AppNavigator/screens/Settings/components/RowLanguageItem.tsx
```diff
@@ -0,0 +1,73 @@
...
+import { getLocaleByLanguageName, translate } from '@translations'
+
+import * as React from 'react'
+import { useCallback } from 'react'
+import { SettingsParamList } from '../SettingsNavigator'
+
+export function RowLanguageItem (props: { language: EnvironmentLanguage }): JSX.Element {
+  const navigation = useNavigation<NavigationProp<SettingsParamList>>()
+  const { language, setLanguage } = useLanguageContext()
+  const onPress = useCallback(async () => {
```
thedoublejay, line 17: don't use useCallback here anymore. It's too buggy on mobile (Like poolswap error). Let's use it only on pure functions

app/screens/AppNavigator/screens/Settings/SettingsScreen.tsx
```diff
@@ -97,27 +97,34 @@ export function SettingsScreen ({ navigation }: Props): JSX.Element {
...
+      <View style={tailwind('mt-4')} />
+      <NavigateItemRow
+        testID='setting_navigate_About'
+        label='About'
+        onPress={() => navigation.navigate('AboutScreen')}
       />
       <RowThemeItem />
+      <NavigateItemRow
+        testID='setting_navigate_language_selection'
+        label='Language'
```
thedoublejay, line 125: This is not translated

thedoublejay (changes requested):

app/screens/AppNavigator/screens/Settings/SettingsScreen.tsx
kyleleow: label prop is translated within `NavigateItemRow`

app/screens/AppNavigator/screens/Settings/components/RowLanguageItem.tsx
kyleleow: @jtlntno-cake from your proto every option is still in english tho, shouldn't the option itself use the language that it means? Like the screenshot i attached above

jtlntno-cake: Reviewed other applications - it makes sense to be aligned with the implementation from the screenshots above. Proto is now updated @kyleleow

thedoublejay (approved):

PR #788, opened 2021-09-06
Activity indicator for submit buttons

### LINKED ISSUE #685: `button` - add activity indicator on buttons
- Add a spinner or loading indicator, e.g.: on the button after user click on "Send" or "Convert" or any form submission button

#### Why is this needed:
- To inform user that the wallet is processing their action

- Confirm Send
- Confirm Convert
- Confirm Pool swap
- Confirm Add liquidity

- Waiting time is noticeable when internet connection is slow

image

fullstackninja864:

/kind feature

Added a spinner or loading indicator, e.g.: on the button after user click on "Send" or "Convert" or any form submission button

Fixes #685

#### Developer Checklist:

- [x] Unit tests*
- [x] Added e2e tests*
- [x] Added translations*


kyleleow (changes requested):
Not sure if it happened too fast, but the loading indicator is not visible when tested on iPhone 12 (refer to video recording below)

Maybe @thedoublejay can help to test on android?

To be extra safe, I suggest maybe adding an e2e test?

https://user-images.githubusercontent.com/40191153/132284411-d2b5de23-3338-428a-be5b-d652b623a573.mov

fullstackninja864:
The activity indicator is when the internet connection is slow, I guess for you its fast, even though we are redirecting user immediately to auth screen when user press send/convert, @thedoublejay , @kyleleow let me know what changes we should do for this

kyleleow:
> The activity indicator is when the internet connection is slow, I guess for you its fast, even though we are redirecting user immediately to auth screen when user press send/convert, @thedoublejay , @kyleleow let me know what changes we should do for this

Is there any way that you used during your development to know that it works? Maybe can add that steps into e2e test.

I did some testing on chrome on your branch, by manually throttling the network speed and adding the value `isSubmitting` flag into the ConvertConfirmationScreen button label, but i am not seeing the flag value being reflected (refer to video below) 

https://user-images.githubusercontent.com/40191153/132305254-82b43d5c-b651-44b0-a53d-74dfb6a7ba94.mov

thedoublejay:
@fullstackninja864 maybe you can upload a video here with the activity indicator

fullstackninja864:
@thedoublejay and @kyleleow  here is the webcast for activity loader

https://user-images.githubusercontent.com/53080940/132451680-9143b7db-a378-44cf-93aa-5b8a72180882.mp4

app/screens/AppNavigator/screens/Balances/screens/SendConfirmationScreen.tsx
```diff
@@ -130,7 +130,7 @@ export function SendConfirmationScreen ({ route }: Props): JSX.Element {
       <SubmitButtonGroup
         isDisabled={isSubmitting || hasPendingJob || hasPendingBroadcastJob}
         label={
-          isSubmitting
+          isSubmitting || hasPendingJob || hasPendingBroadcastJob
```
kyleleow, line 133: Do correct me if I am wrong, will there be a scenario where we have pending job from other transaction but we want to display a contextual loading message here? 

Yep, it would be useful if that job happen to be "Sending", but it can also be "Converting" or "Adding" no? In my opinion, I think by disabling the button is sufficient for this case

Similarly for all the other <SubmitButtonGroup> implementation

kyleleow: Also, would it be more readable if we add a new prop that represent loading label explicitly? And another prop that indicate the loading state?

This is a question, it may be difficult to implement I don't know, happy to discuss

fullstackninja864: In my opinion Loader should be visible only when `isSubmitting ` is true, But as we are dispatching txn into queue and just updating the state of `isSubmitting` to false, we are getting some delay in opening the confirmation screen and turning off Loader, Thats why I have added `isSubmitting || hasPendingJob || hasPendingBroadcastJob` items too for the loader, though As you mentioned if suppose we initiated the txn and went for second one then before confirmation screen we have one screen where we fill the values of txn (For eg SendComponent) there we are not allowing user to press Send button until previous txn is in progress (using same condition) https://github.com/DeFiCh/wallet/blob/main/app/screens/AppNavigator/screens/Balances/screens/SendScreen.tsx#L62. so user wont be able to visit confirm screen until old txn is not get submitted/confirmed
```
async function onSubmit (): Promise<void> {
    if (hasPendingJob || hasPendingBroadcastJob) {
      return
    }
```

app/components/themed/ThemedTextActivityIndicator.tsx
```diff
@@ -0,0 +1,29 @@
+import React from 'react'
+import { View } from 'react-native'
+import { tailwind } from '@tailwind'
+import { ThemedActivityIndicator } from './ThemedActivityIndicator'
+import { ThemedText } from './ThemedText'
+import { ThemedProps } from '.'
+
```
thedoublejay, line 7: I prefer adding it on the button itself. No need to have this element since it's always displayed in a button. It would be better to just pass a "loading" param in the button (that triggers the display for ActivityIndicator) and dynamic text (e.g, converting.. convert etc.)

thedoublejay (changes requested):

thedoublejay (changes requested):
Keep all labels (submitting and non-submitting) uppercase. Example, it changes between "SEND" to "Sending".

app/screens/AppNavigator/screens/Dex/DexConfirmAddLiquidity.tsx
```diff
@@ -171,6 +170,8 @@ export function ConfirmAddLiquidityScreen (props: Props): JSX.Element {
       <SubmitButtonGroup
         isDisabled={isSubmitting || hasPendingJob || hasPendingBroadcastJob}
         label={translate('screens/ConfirmAddLiq', 'ADD')}
+        isSubmitting={isSubmitting || hasPendingJob || hasPendingBroadcastJob}
+        submittingLabel={translate('screens/ConfirmAddLiq', 'Adding Liquidity')}
```
thedoublejay, line 174: Just use ADDING here, to prevent overflow of long texts

thedoublejay (changes requested):
Same for other operations, just use the progressive tenses (ADDING, SENDING etc.)

app/screens/AppNavigator/screens/Dex/DexConfirmAddLiquidity.tsx
fullstackninja864: Will be updating `Removing Liquidity` to `REMOVING` and `Swapping Token` to `SWAPPING`

thedoublejay (approved):

kyleleow (approved):

PR #832, opened 2021-09-10
added tsc in lint check

### LINKED ISSUE #428: lint workflow should run tsc to check TypeScript types

In `ci.yml` workflow or `scripts: lint` we should check TypeScript types.

#### Why is this needed:

The current lint and compile are babel-based, it doesn't check types as you normally would expect a TS project to do. We need it to check for compile type errors. 

/area core workflow
/triage accepted

thedoublejay:

/kind chore

Fixes #428

#### Developer Checklist:

- [ ] Unit tests*
- [ ] Added e2e tests*
- [ ] Added translations*

### DIFF
--- a/tsconfig.production.json
+++ b/tsconfig.production.json
@@ -1,0 +1,7 @@
+{
+  "extends": "./tsconfig.json",
+  "exclude": [
+    "node_modules",
+    "cypress"
+  ]
+}
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,5 +1,4 @@
 {
-  "root": true,
   "parserOptions": {
     "project": "./tsconfig.json"
   },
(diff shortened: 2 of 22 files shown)


tsconfig.production.json
```diff
@@ -0,0 +1,7 @@
+{
+  "extends": "./tsconfig.json",
+  "exclude": [
+    "node_modules",
+    "cypress"
```
fuxingloh, line 5: awkward just for this line due to eslint

fuxingloh (approved):

tsconfig.production.json
thedoublejay: yeah, I tried many options but it breaks due to eslint dependency. Might need to optimize soon but so far this works.

PR #840, opened 2021-09-11
disabled poolswap button - remove max amount for token B

### LINKED ISSUE #689: `poolswap` - DFI - BTC disabled on some cases

### LINKED ISSUE #728: cannot swap DFI to BTC from bigger amount of 1 DFI

thedoublejay:

/kind fix

Currently, when doing pool swaps on mainnet. There are some cases where "continue" button is disabled. This is due to the fact that we're using a "max amount" for Token B. We have this behavior before because we're allowing input in TokenB (currently, it's disabled). 

There's a slight difference on the calculated amount and max amount causing it to be invalid. Since we're not allowing editing of Token B, we don't need maxAmount for it.

Fixes #728 
Fixes #689 

#### Developer Checklist:

- [x] Unit tests*
- [x] Added e2e tests*
- [x] Added translations*

### DIFF
--- a/app/screens/AppNavigator/screens/Dex/PoolSwap/PoolSwapScreen.tsx
+++ b/app/screens/AppNavigator/screens/Dex/PoolSwap/PoolSwapScreen.tsx
@@ -190,7 +190,6 @@ export function PoolSwapScreen ({ route }: Props): JSX.Element {
         control={control}
         controlName={tokenBForm}
         isDisabled
-        maxAmount={aToBPrice.times(getValues()[tokenAForm]).toFixed(8)}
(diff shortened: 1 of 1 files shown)


fuxingloh (approved):
oh the rules

PR #918, opened 2021-09-21
Basic push notifications implementation for txn confirmation and failure update

### LINKED ISSUE #644: push notifications
Allows us to send notifications (if user permits). Ex. when a transaction has been confirmed, etc.

The scenario would be, assuming users allow notifications:
1. Send a token
2. Transaction signed and broadcasted
3. Wait for it to be sent/timeout 
4. Send notification

Some things to consider:

- [x] Only push when allowed
- [x] Translations 
- [x] Add logo of app
- [x] Only push notification when user still has the same wallet loaded
- [x] Only push notification if wallet is on background

#### Why is this needed:
https://docs.expo.dev/versions/latest/sdk/notifications/

fullstackninja864:

/kind feature

In this PR we have implemented in-app basic push notification flow. The push notification will work when app will be in foreground.

The lifecycle of push notification is as follows

1.  When app loads we are fetching push notification permission of the app and storing it in the  `status` variable for future use.
2. If we don't have any permission then we are requesting for it, from the same `request` function.
3. Currently the push notifications are implemented for success and failure of Txn if we have notification permission status as `GRANTED`.  From a common Txn initiation point (OceanInterface component) we are initiating success or error notification.

Fixes #644

#### Developer Checklist:

- [x] Unit tests*
- [x] Added e2e tests*
- [x] Added translations*


app/screens/RootNavigator.tsx
```diff
@@ -13,6 +14,10 @@ import { WalletNavigator } from './WalletNavigator/WalletNavigator'
 export function RootNavigator (): JSX.Element {
   const { wallets, isLoaded } = useWalletPersistenceContext()
 
+  useEffect(() => {
+    void WalletNotifications.register()
```
thedoublejay, line 18: why is there a void here?

app/screens/RootNavigator.tsx
```diff
@@ -1,4 +1,5 @@
-import React from 'react'
+import { WalletNotifications } from './../api/wallet/notifications'
```
thedoublejay, line 1: shortcut

app/screens/TransactionAuthorization.tsx
```diff
@@ -184,6 +185,12 @@ export function TransactionAuthorization (): JSX.Element | null {
         .then(async signedTx => {
           // case 1: success
           await resetPasscodeCounter()
+          if (transaction?.title != null && transaction?.description != null) {
```
thedoublejay, line 188: Better move it to `OceanInterface`. So we can trigger it during:
1. Before broadcast
2. On completion of broadcast
3. If there are any errors

thedoublejay (changes requested):

app/screens/RootNavigator.tsx
fullstackninja864: TO avoid floating promise warning we have added void

app/screens/TransactionAuthorization.tsx
```diff
@@ -28,6 +28,7 @@ import { ocean } from '@store/ocean'
 import { DfTxSigner, first, transactionQueue } from '@store/transaction_queue'
 import { tailwind } from '@tailwind'
 import { translate } from '@translations'
+import { WalletNotifications } from '../api/wallet/notifications'
```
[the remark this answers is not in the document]
fullstackninja864: Will get updated in next push

app/screens/RootNavigator.tsx
fullstackninja864: Will get updated in next push

app/components/OceanInterface/OceanInterface.tsx
```diff
@@ -130,6 +132,12 @@ export function OceanInterface (): JSX.Element | null {
             Logging.error(e)
             title = 'Sent but not confirmed'
           }
+
+          notificationObj = {
+            title,
```
thedoublejay, line 137: Need to translate this

app/components/OceanInterface/OceanInterface.tsx
```diff
@@ -130,6 +132,12 @@ export function OceanInterface (): JSX.Element | null {
...
+            body: `Txid: ${transaction.tx.txId}`
```
thedoublejay, line 138:
```suggestion
            body: `Transaction ID: ${transaction.tx.txId}`
```

Need to translate `Transaction ID` also

thedoublejay (changes requested):

thedoublejay (dismissed):

mobile-app/app/components/OceanInterface/OceanInterface.tsx
```diff
@@ -138,10 +146,18 @@ export function OceanInterface (): JSX.Element | null {
...
+            title: translate('screens/OceanInterface', 'Transaction Error'),
+            body: translate('screens/OceanInterface', err.message)
+          }
           setError(errMsg)
           Logging.error(e)
         })
         .finally(() => {
+          if ((notificationObj.title !== '') && (notificationObj.body !== '')) {
+            void WalletNotifications.send(notificationObj)
+          }
```
fuxingloh, line 160: no point separating the cause and effect?

You split them up into the assignment of object, setting of object, and dispatching which is redundant IMO

You could just dispatch them via `void WalletNotifications.send({...})` for more simple implementation and lower cognitive complexity

mobile-app/app/api/wallet/notifications.ts
```diff
@@ -0,0 +1,77 @@
...
+  title: string
+  body: string
+  data?: { [key: string]: unknown }
+}
+
+/**
+ * Wallet push notification implementation light wallet
+ */
+class NotificationsService {
+  status: PermissionStatus | undefined
```
fuxingloh, line 15:
```suggestion
  status?: PermissionStatus
```

mobile-app/app/api/wallet/notifications.ts
```diff
@@ -0,0 +1,77 @@
+import * as Notifications from 'expo-notifications'
+import { Platform } from 'react-native'
+import { PermissionStatus } from 'expo-modules-core'
+
+export interface SendNotificationProps {
+  title: string
+  body: string
+  data?: { [key: string]: unknown }
```
fuxingloh, line 8: avoid `unknown` or ` any`

mobile-app/app/screens/RootNavigator.tsx
```diff
@@ -16,6 +17,10 @@ export function RootNavigator (): JSX.Element {
     isLoaded
   } = useWalletPersistenceContext()
 
+  useEffect(() => {
+    void WalletNotifications.register()
+  }, [])
```
fuxingloh, line 22: this should be part of settings and asked upon wallet creation since you are asking for permission, should not register every time?

mobile-app/app/api/wallet/notifications.ts
```diff
@@ -0,0 +1,77 @@
+import * as Notifications from 'expo-notifications'
+import { Platform } from 'react-native'
+import { PermissionStatus } from 'expo-modules-core'
+
+export interface SendNotificationProps {
```
fuxingloh, line 5:
```suggestion
export interface SendNotificationData {
```

avoid use of Props for non React common to reduce confusion

mobile-app/app/api/wallet/notifications.ts
```diff
@@ -0,0 +1,77 @@
...
+          body,
+          data
+        },
+        trigger: null
+      })
+    }
+  }
+}
+
+export const WalletNotifications = new NotificationsService()
```
fuxingloh, line 77: it shouldn't be a class, should just be plain old functions?

fuxingloh (changes requested):
the lifecycle of notification needs some thinking and UX considerations 

1. Registering Notification on? e.g. Wallet Creation, Settings?
2. Settings to change Notification Delivery Setting
3. Notification copy should highlight it only works for xxx scenario, e.g when it's in background mode. This should be prompted somewhere.

retitle your PR to something more meaning, "push notification" is too broad.

mobile-app/app/api/wallet/notifications.ts
fullstackninja864: When our app loads, we first check the push notification permission using [register](https://github.com/DeFiCh/wallet/blob/a6ce2394b345a3df3234a0320e4d4940f5c02078/mobile-app/app/api/wallet/notifications.ts#L20) function and set the permission status to `status` variable, and before sending the notification we first check the same status is granted or not ([link](https://github.com/DeFiCh/wallet/blob/a6ce2394b345a3df3234a0320e4d4940f5c02078/mobile-app/app/api/wallet/notifications.ts#L64)). For storing the status, it's better we use class function, otherwise we have to store this in context or in redux store or in localStore.

mobile-app/app/api/wallet/notifications.ts
fullstackninja864: Will do the needful

mobile-app/app/screens/RootNavigator.tsx
fullstackninja864: With the register function we are first fetching notification permission, and if the status is `GRANTED` then we are not requesting it again but just updating the [status](https://github.com/DeFiCh/wallet/blob/a6ce2394b345a3df3234a0320e4d4940f5c02078/mobile-app/app/api/wallet/notifications.ts#L15) variable for future purpose

mobile-app/app/api/wallet/notifications.ts
fullstackninja864: will do the changes as suggested

mobile-app/app/components/OceanInterface/OceanInterface.tsx
fullstackninja864: will do the changes as suggested

mobile-app/app/api/wallet/notifications.ts
fullstackninja864: will update the changes

thedoublejay:
I'll loop in @jtlntno-cake with this:

As Fuxing pointed, we would like to have some kind of lifecycle for notifications:

1. No need for auto-register. Enable it on Settings page. (Push Notifications or something)
2. We need to explain how it works to users and what are available to them. I suggest checking the Instagram notification page. We can have a dedicated screen that displays all notification types (e.g, Transaction completion etc.). 
3. Status of the notification should also be visible
  - [ ] Not active but enabled on system level
  - [ ] Active and enabled on system level
  - [ ] Inactive and disabled on system level
  - [ ] Active and disabled on system level

CC: @fullstackninja864

jtlntno-cake:
Hey guys, please see [prototype](https://www.figma.com/proto/SDIpZciuxJlUf4xf8yheiY/Mobile-Light-Wallet?node-id=287%3A11410&viewport=241%2C48%2C1.88&scaling=scale-down&starting-point-node-id=841%3A28270&show-proto-sidebar=1) for toggling push-notifications. It will be a page for toggling the notifications as we may have to scale this feature in the future. 

The screen can be seen within this [Figma board area](https://www.figma.com/file/SDIpZciuxJlUf4xf8yheiY/Mobile-Light-Wallet?node-id=1470%3A40591)

Also, we use native design for push notifications. Please see this [link](https://www.figma.com/file/XdKJSI9TGYAf1VXJCQf2cZ/DeFiChain-Design-System?node-id=606%3A6596) if needed some reference

mobile-app/app/api/persistence/notifiction_storage.ts
```diff
@@ -0,0 +1,52 @@
...
+      displayName: 'Transactions',
+      dafaultValue: true
+    }
+  ]
+}
+
+function notificationParser (preferances: AppNotificationTypesI[]): AppNotificationTypesI[] {
+  const notificationsTypes = getAppNotificationTypes()
+  return notificationsTypes.map(({ type, displayName, dafaultValue }) => {
+    const savedType = preferances.find((each) => {
```
thedoublejay, line 32: Can we just save the preference type? So we don't need to do double loop (map and find).

mobile-app/app/screens/AppNavigator/screens/Settings/screens/NotificationScreen.tsx
```diff
@@ -0,0 +1,88 @@
...
+
+  useEffect(() => {
+    NotificationPersistence.get()
+      .then((notificationsT) => {
+        setNotificationsTypes(notificationsT)
+      })
+      .catch((err) => Logging.error(err))
+  }, [])
+
+  const onPreferanceChange = async (type: NotificationType, value: boolean): Promise<void> => {
```
thedoublejay, line 24: preference

mobile-app/app/screens/RootNavigator.tsx
thedoublejay: If we set notifs as false by default, can we just call it when we enable it?

thedoublejay (changes requested):

thedoublejay:
@fullstackninja864 Notifs are not pushed if app is on background.

fullstackninja864:
@thedoublejay , @kyleleow 
1. Related to transaction context it was discussed with JJ that for now we only send success or failure notification.
2. Also, as discussed in our previous call that we decided to lay a base foundation for notification.
3. There are technical limitations when sending notification when app is running in background, reason being the way we implemented the setTimeout for transaction polling, when app is in background it setTimeout get's paused.
4. As discussed with JJ we are not using sound notification.

@thedoublejay  I think we should update the ticket as it doesn't reflect our last conversations and decisions.

thedoublejay:
Will be closing this for now, will re-open once we work on it again

PR #954, opened 2021-09-28
Blocks height link for defi scan

### LINKED ISSUE #948: `network details` - block height details
In network details, we show the block height. We should make it clickable and redirect to DeFiScan block details page.

https://defiscan.live/blocks/1228320

fullstackninja864:

/kind feature

Make block height clickable and redirect to DeFiScan block details page.

Fixes #948

#### Developer Checklist:

- [x] Unit tests*
- [x] Added e2e tests*
- [x] Added translations*


fullstackninja864:
SS of implementation 
image

thedoublejay (changes requested):
Make the block number pink, check Transaction URL in Ocean Interface as reference.

kyleleow:
Since it is clickable, might want to add an e2e test to it?

fullstackninja864:
@kyleleow  , Thanks for suggestion, I have added e2e for the block height link.

mobile-app/app/contexts/DeFiScanContext.tsx
```diff
@@ -58,3 +65,8 @@ function getTxURLByNetwork (network: EnvironmentNetwork, txid: string, rawtx?: s
 
   return baseUrl
 }
+
+function getBlocksURLByNetwork (network: EnvironmentNetwork, blockCount: number): string {
+  const baseUrl = `https://defiscan.live/blocks/${blockCount}${getNetworkParams(network)}`
```
thedoublejay, line 70: Can we store the URL as a constant on this file?

thedoublejay (changes requested):

mobile-app/app/contexts/DeFiScanContext.tsx
thedoublejay: URL is used on getBlocksURL and getTxURL

fullstackninja864: I will do that in next PR for tokens where we will use a common method to get the url

thedoublejay (approved):

PR #981, opened 2021-09-30
trigger EAS build on merge to `main branch`

thedoublejay:

/kind chore

On successful merge to main branch, this will "queue" a build to EAS build. Still needs to be approved. 

Fixes #

#### Developer Checklist:

- [ ] Unit tests*
- [ ] Added e2e tests*
- [ ] Added translations*


.github/workflows/eas-build.yml
```diff
@@ -0,0 +1,57 @@
...
+      - uses: actions/github-script@441359b1a30438de65712c2fbca0abe4816fa667
+        id: version
+        with:
+          script: |
+            const semver = '${{steps.previoustag.outputs.tag}}'.replace('v', '')
+            if (semver.match(/^[0-9]+\.[0-9]+\.[0-9]+$/)) {
+              return semver
+            }
+
+            throw new Error('not semver')
```
fuxingloh, line 35: how about using time e.g. `0.0.2109301845` should we know the exact build time?

fuxingloh (changes requested):

.github/workflows/eas-build.yml
fuxingloh: Also not sure how the App Store or OS will react if there is duplicated versioning with the current method.

thedoublejay: For APK, need to uninstall the current app and install the new one. I think should be ok since it's just for testing purposes.

I'll try the build number if it will be accepted by standard-version

.github/workflows/eas-build.yml
```diff
@@ -0,0 +1,57 @@
...
+          result-encoding: string
```
fuxingloh, line 36:
```suggestion
      - uses: actions/checkout@5a4ac9002d0be2fb38bd78e4b4dbde5606d7042f

      - uses: actions/github-script@441359b1a30438de65712c2fbca0abe4816fa667
        id: version
        with:
          script: |
            const date = new Date()
            const build = '0.0.' + date.toISOString().replace(/[^0-9]/g, '')
            console.log(build)
            return build

          result-encoding: string
```

.github/workflows/eas-build.yml
```diff
@@ -0,0 +1,51 @@
...
+      - uses: actions/checkout@5a4ac9002d0be2fb38bd78e4b4dbde5606d7042f
+
+      - uses: actions/github-script@441359b1a30438de65712c2fbca0abe4816fa667
+        id: version
+        with:
+          script: |
+            const date = new Date()
+            const build = '0.0.' + date.toISOString().replace(/[^0-9]/g, '')
+            console.log(build)
+            return build
```
fuxingloh, line 28:
```suggestion
            const [date] = new Date().toISOString().split('T')
            const build = '0.0.' + date.replace(/[^0-9]/g, '')
            console.log(build)
            return build
```

fuxingloh (approved):

PR #983, opened 2021-09-30
Wait for Privacy lock and Storage to load

thedoublejay:

/kind fix

Currently, local storage setting (Privacy Toggle) and Hardware Async are checked before we display the prompt. This fix waits for the two promises before displaying the contents. Needs to be fully tested on a standalone device.

Fixes #

#### Developer Checklist:

- [ ] Unit tests*
- [ ] Added e2e tests*
- [ ] Added translations*


kyleleow (approved):

mobile-app/app/contexts/LocalAuthContext.tsx
```diff
@@ -64,6 +68,9 @@ export function PrivacyLockContextProvider (props: React.PropsWithChildren<any>)
         Logging.error(error)
         setIsPrivacyLock(false)
       })
+      .finally(() => {
+        setIsDevicePersistenceLoaded(true)
+      })
   }, [/* only load from persistence layer once */])
```
kyleleow, line 74: do you need to add any dependency here so that we can rerender again?

PR #1211, opened 2021-11-02
Feature gate

### LINKED ISSUE #725: Feature Gating

Feature gating creates a list of configurations isolated from the main app, accessible via settings. They can be used to enable/disable app features.

With feature gating, we can release new features for testing directly into the main app and have it disabled. Advanced users can choose to enable features while they are in beta.

#### Why is this needed:

This allows us to move faster bring testing features early to a handful of users. Release early, test early, and iterative early.

fullstackninja864:

/kind feature

In this PR we are allowing user to opt for  Beta features using  Feature gating of configurations isolated from the main app, accessible via settings > About > Try Beta features . They can be used to enable/disable app features.
With feature gating, we can release new features for testing directly into the main app and have it disabled. Advanced users can choose to enable features while they are in beta.

To use the Feature Gate we have two ways available
1. Using `FeatureGate` HOC
```
<FeatureGate feature='<featureId>'>
  <ChildComponent />
</FeatureGate>
```
2. using `isFeatureAvailable` method
```
const isAvailable = isFeatureAvailable('<featureId>')
```
also when user enable the feature, then in order to show users warning related to risk involved using beta feature we can use `isBetaFeature` method to get boolean value.

Fixes #725

#### Developer Checklist:

- [x] Unit tests*
- [ ] Added e2e tests*
- [x] Added translations*


mobile-app/app/screens/AppNavigator/screens/Settings/screens/FeatureFlagScreen.tsx
```diff
@@ -0,0 +1,130 @@
...
+  value: boolean
+}
+
+export function FeatureFlagScreen (): JSX.Element {
+  const { featureFlags, enabledFeatures, updateEnabledFeatures } = useFeatureFlagContext()
+  const features = featureFlags.filter((item) => item.stage === 'beta')
+  const [betaFeatures, setBetaFeatures] = useState<BetaFeaturesI []>([])
+
+  const getBetaFeature = (flags: FEATURE_FLAG_ID[]): BetaFeaturesI[] => {
+    return features.map((item: FeatureFlag) => {
```
thedoublejay, line 22: we are looping here twice (loop on another loop). Maybe we can change it to a better option

thedoublejay (changes requested):
1. Try Beta Feature button has an underline
2. It's too close to components above and below it, maybe add some margin

website/src/pages/api/v0/settings/flags.ts
```diff
@@ -11,8 +11,9 @@ export default async function handle (req: NextApiRequest, res: NextApiResponse<
   await runMiddleware(req, res, cors)
   res.json([{
     id: 'loan',
-    name: 'Decentralized Loans',
-    stage: 'alpha',
-    version: '>=0.12.0'
+    name: 'Loans',
+    stage: 'beta',
```
thedoublejay, line 15: @fullstackninja864 this should be Alpha, don't change it to Beta yet.

thedoublejay (changes requested):

website/src/pages/api/v0/settings/flags.ts
fullstackninja864: Sure, will update it

mobile-app/app/screens/AppNavigator/screens/Settings/screens/AboutScreen.tsx
```diff
@@ -103,6 +105,10 @@ const SOCIAL_LINKS: AboutScreenSocialLinks[] = [
 ]
 
 export function AboutScreen (): JSX.Element {
+  const navigation = useNavigation<NavigationProp<SettingsParamList>>()
+  const { featureFlags } = useFeatureFlagContext()
+  const features = featureFlags.filter((item) => item.stage === 'beta')
```
kyleleow, line 110: `featureFlags.some()` could be a better option here if you're only interested if an item exist

kyleleow (changes requested):
Can i request to add some e2e test cases? We can stub the `getFeatureFlag()` API to simulate different scenario

mobile-app/app/contexts/FeatureFlagContext.tsx
```diff
@@ -20,51 +25,76 @@ export function useFeatureFlagContext (): FeatureFlagContextI {
...
-  const [isAuctionDisplayed, setIsAuctionDisplayed] = useState<boolean>(false)
+  const [enabledFeatures, setEnabledFeatures] = useState<FEATURE_FLAG_ID[]>([])
+
   if (!isError) {
     prefetchPage({})
   }
 
+  function isBetaFeature (featureId: FEATURE_FLAG_ID): boolean {
+    return featureFlags.some((flag) => {
+      return flag.id === featureId && flag.stage === 'beta'
```
kyleleow, line 44: need to check for app version if its mobile

mobile-app/app/screens/AppNavigator/screens/Settings/screens/AboutScreen.tsx
fullstackninja864: Will update this, Thanks for suggestions

kyleleow (approved):

thedoublejay (approved):

PR #1213, opened 2021-11-02
`desktop` - DeFiChain Wallet

### LINKED ISSUE #864: Create desktop structure with Electron

thedoublejay:

/kind feature

Creates initial desktop structure for Electron. 
1. Added `electron-build.yml` to trigger dev builds on PR and push to main
   - Notarization step for Mac OS
2. To run the app, go to `desktop-app` directory and run `npm run start:dev`
   - It will run expo local server and trigger initiate-electron.js
   - initiate-electron.js will wait for the expo server to be available before creating the window
3. Added basic menu for Window
4. On dev, it will run with Playground screen. 
5. If you want to test out a binary, you can run a specific command for your OS. 
  - example: `npm run build:win` - will trigger a binary build for Windows. It will be on your `desktop-app/dist`
6. Fixed `linking.ts` to open web to new tabs instead of same app
7. Added `index.html` and `webpack.config.js` from @expo/config to customize paths for Web and Desktop. Reference - https://docs.expo.dev/guides/customizing-webpack/

Fixes #864 
Fixes https://github.com/DeFiCh/wallet/issues/865
Fixes https://github.com/DeFiCh/wallet/issues/867 (Mac)

#### Developer Checklist:

- [ ] Unit tests*
- [ ] Added e2e tests*
- [ ] Added translations*


mobile-app/app/screens/WalletNavigator/screens/CreateWallet/OnboardingNetworkSelectScreen.tsx
```diff
@@ -4,10 +4,10 @@ import { RowNetworkItem } from '@components/RowNetworkItem'
 import { translate } from '@translations'
 import * as React from 'react'
 import { View } from 'react-native'
-import * as Updates from 'expo-updates'
+import { getReleaseChannel } from '@api/releaseChannel'
```
fuxingloh, line 7: We should separate out the release channel changes first

fuxingloh (changes requested):

mobile-app/app/screens/WalletNavigator/screens/CreateWallet/OnboardingNetworkSelectScreen.tsx
thedoublejay: I added it because release channel is only updated on eas builds (so all electron builds are considered as dev builds).

thedoublejay: @fuxingloh ok I have separated the release channel commit - https://github.com/DeFiCh/wallet/pull/1224

fuxingloh: could the PR be updated to target the main first, I am thinking about whether it can be refactored into a Context for inversion of control. Provided once and injected everywhere, now we are pulling it causing complexity in dependency injecting

thedoublejay: Oh damn I have my commits there haha wait let me create another one

thedoublejay: @fuxingloh updated - https://github.com/DeFiCh/wallet/pull/1229/files

thedoublejay: Hmmm we can do that also, something like Environment Context

thedoublejay: Let me check

fuxingloh: I am also very curious about the thoughts of using Context via the `Provider` and `Consumer` patterns in this project. It's certainly some level of implicit injection to it. I always believe it helps the project in the long run where you don't need to worry managing dependency and it's outsourced to the context and provided where applicable.

ivan-zynesis: Generally agree that it should be beneficial. Some hidden "cons" may need extra awareness

- redundant workload introduced when the context is not even used more than once
- when context is available but not being used effeciently

IMO these API injections are not needed can be embedded within the provider. Provider should be able to construct them using the `useSecureStorage`
image

thedoublejay: @ivan-zynesis I agree, it's just that some of our functions (e.g. secured_storage, etc.)  are not yet context (therefore, cannot use hooks at the meantime). So we don't really have a choice but to pass it as props (if we want to re-use them). It will be cleaner once we have refactored most of it to be context.

fuxingloh:
Looking at this PR now

PR #1274, opened 2021-11-11
Address Switching - Generating and using new Address in Light Wallet

### LINKED ISSUE #910: Address Switching - Generating and using new Address in Light Wallet

Currently, on the Light Wallet, only `index:0` addresses are used. The internal implementation of `jellyfish-wallet` allows multi addresses to be used. This issue calls for implementing wallet management with multi addresses supported. 

1. Listing all the active addresses within the wallet.
2. Generating a new Address `n+1` and storing the max `index` to know how many addresses have been generated.
3. Discovering all addresses on the wallet via https://github.com/DeFiCh/jellyfish/blob/47c3dabe7995bbfe8e7812ea1dfd1e885a6355f7/packages/jellyfish-wallet/src/wallet.ts#L69-L77 and storing the max indexes.

Assumptions:
- We are not using Node-like address management where all addresses are pooled together with one balance, for simplicity reasons each address is an account within the wallet system that the user can switch to.
- We are following a Metamask like experience
- We are not going to store all addresses like a wallet.dat file, we just need to track where the max index is.

Max Index Length determines the maximum number of activated addresses allowed in the wallet. Accounts are considered active if the address contains any transaction activity. You can always generate an address (+1). You can't generate another address until you have activated the previous address. You active a previous address by using it. (contains any transaction activity)

- `Max Index Length = 1`, only 1 default address
- `Max Index Length = 2`, 2 addresses [Active, Active/In-Inactive]
- `Max Index Length = 3`, 3 addresses [Active, Active, Active/In-Inactive]

image

* Avatars of wallet are already implemented.

/area core wallet

fullstackninja864:

/kind feature

Fixes #910

#### Developer Checklist:

- [ ] Unit tests*
- [x] Added e2e tests*
- [x] Added translations*


thedoublejay (changes requested):
One super early feedback would be to check all usage of `wallet.get(0)`. Currently, we're calling it https://github.com/DeFiCh/wallet/blob/db648f7b8e02e3e60f0964138d21063c90acbc67/mobile-app/app/screens/TransactionAuthorization/TransactionAuthorization.tsx#L222 

I'm not sure if this will be affected, but good to check.

thedoublejay:
@jtlntno-cake One thing we're missing from the design is, if you open the address switcher, we don't have an indicator which address is currently active/selected.

jtlntno-cake:
@thedoublejay Adding this [tag in the design](https://www.figma.com/file/SDIpZciuxJlUf4xf8yheiY/PRDCT---Mobile-Light-Wallet?node-id=3574%3A52767) to indicate which wallet is currently used by user

Screen Shot 2021-11-12 at 10 57 15 AM

shared/contexts/WalletContext.tsx
```diff
@@ -40,20 +63,53 @@ export function WalletContextProvider (props: PropsWithChildren<any>): JSX.Eleme
   }, [provider, network, client])
 
   useEffect(() => {
-    wallet.get(0).getAddress().then(value => {
-      setAddress(value)
-    })
+    getWalletDetails()
```
fuxingloh, line 66: https://github.com/DeFiCh/jellyfish/blob/main/packages/jellyfish-wallet/src/wallet.ts#L77

You need to discover the `addresses` automatically when the user recovers their wallet and set the max/length.

Not here but somewhere in the recover wallet screens.

fuxingloh (commented):
Some suggestions

mobile-app/app/screens/AppNavigator/screens/Balances/components/AddressControlScreen.tsx
```diff
@@ -0,0 +1,168 @@
...
+  )
+}
+
+export function AddressControlCard ({ onClose }: { onClose?: () => void }): JSX.Element {
+  const { availableAddresses, switchAccount, createAddress, wallet } = useWalletContext()
+  const [canCreateAddress, setCanCreateAddress] = useState<boolean>(false)
+  const logger = useLogger()
+
+  useEffect(() => {
+    wallet.discover(availableAddresses.length).then(discoveredAddress => {
```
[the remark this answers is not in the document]
fullstackninja864: We will need this discover at two place
1. when user restore the wallet 
2. when user try to create new address. As mentioned in ticket
Max Index Length = 1, only 1 default address
Max Index Length = 2, 2 addresses [Active, Active/In-Inactive]
Max Index Length = 3, 3 addresses [Active, Active, Active/In-Inactive]
You can't generate another address until you have activated the previous address. To get this info we need discover here too

mobile-app/app/api/wallet/address_index.ts
```diff
@@ -0,0 +1,23 @@
+import { SecuredStoreAPI } from '@api'
+
+const KEY = 'WALLET_ADDRESS.INDEX'
+
+export type AddressIndexType = 'max' | 'active'
```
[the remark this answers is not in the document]
fullstackninja864: Will update this

mobile-app/app/tailwind.config.js
```diff
@@ -117,6 +121,11 @@ module.exports = {
       },
       zIndex: {
         '-1': '-1'
+      },
+      fontSize: {
```
thedoublejay, line 125: I prefer not introducing a new font for this. Just use text-xs for the "Active" tag. and maybe increase the size of Address text

mobile-app/app/tailwind.config.js
```diff
@@ -108,6 +108,10 @@ module.exports = {
           500: '#FF9F0A',
           600: '#FFBC54',
           700: '#BF7708'
+        },
+        darkblue: {
```
thedoublejay, line 112: we have a different shading for the colors. Please refer to this: https://www.figma.com/file/XdKJSI9TGYAf1VXJCQf2cZ/DSGN---Foundations?node-id=3%3A360

We can use `blue` and `darkblue` for the shades.


```
blue: {
  50: '#F7FBFE',
  100: '#EFF6FD',
  200: '#D6EAF9',
  300: '#BDDDF5',
  500: '#5AA9E6',
  600: '#5198CF',
  700: '#447FAD',
  900: '#2C5371'
},


darkblue: {
  50: '#242D33',
  100: '#3E627A',
  200: '#4C7895',
  300: '#5F96BA',
  500: '#7FC8F8',
  600: '#A5D9FA',
  700: '#CCE9FC',
  900: '#F2FAFE'
},
```

thedoublejay (changes requested):

mobile-app/app/screens/AppNavigator/screens/Balances/components/AddressControlScreen.tsx
```diff
@@ -0,0 +1,190 @@
...
+            dark={tailwind('bg-darkblue-100')}
+            style={tailwind('h-4 ml-1')}
+            testID='address_active_indicator'
+          >
+            <ThemedText
+              light={tailwind('text-blue-500')}
+              dark={tailwind('text-darkblue-500')}
+              style={tailwind('text-2xs px-1 font-medium')}
+            >
+              {translate('screens/AddressControlScreen', 'ACTIVE')}
```
thedoublejay, line 174: Need translations for this. Also, it's not centered properly on the box

thedoublejay (changes requested):

thedoublejay (changes requested):
@fullstackninja864 Some findings after initial testing:

1. Create a new address > Exit Wallet > Restore wallet > If address is not active, it will not be displayed.
2. Avatars are too small on the bottom sheet. Please check with design.
3. Active still not centered, Aligned on the bottom
4. Create wallet address has a different background on dark mode
5. Rectangle above bottom sheet has a different color in the design
6. Need a way to display total number of addresses. @jtlntno-cake has a design for this but not sure if it's the clear. @jtlntno-cake can you verify if this is the expected behavior? The notif will display how many addresses the user has.
image
7. Showing the bottom modal takes long time to display (S9+, around 3-4 seconds to open)
8. There's an error (setState issue, non-breaking) when selecting an address (see console log)
9. On start up, there's a small time where the first address is always the active address despite of selecting a different one. Displaying it for some time then flicker to the actual selected address. This will be a potential issue since user can do a transaction on this period of time.
10. Create new address button is displayed after some time when the bottom sheet is already open

thedoublejay (approved):

thedoublejay (dismissed):

mobile-app/app/screens/AppNavigator/screens/Balances/components/AddressControlScreen.tsx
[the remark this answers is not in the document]
fuxingloh: You only discover a wallet when you restore a wallet.

When you attempt to create a new address, you get the index of address you want to create and check with `wallet.isUsable(index)`.

https://github.com/DeFiCh/jellyfish/blob/97db41f52a845bc90b5a9d249d59ef309bd92831/packages/jellyfish-wallet/src/wallet.ts#L47-L61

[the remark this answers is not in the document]
fuxingloh: This PR is incomplete until this is implemented, we can't discover the wallet at each context load, it's an expensive operation. It's a pretty bad UX due to the extended startup time, further extended if users have many addresses.

mobile-app/app/screens/AppNavigator/screens/Balances/components/AddressControlScreen.tsx
```diff
@@ -0,0 +1,192 @@
...
+    setCanCreateAddress(addresses.length === discoveredAddress.length)
+  }
+
+  const onRowPress = async (index: number): Promise<void> => {
+    await setIndex(index)
+    onClose()
+  }
+
+  useEffect(() => {
+    discoverAddresses().catch(logger.error)
```
fuxingloh, line 89: Should set loading indicator for this since it's an expensive and long operation.

fuxingloh (changes requested):
https://github.com/DeFiCh/wallet/pull/1274#discussion_r749910217

mobile-app/app/screens/AppNavigator/screens/Balances/components/AddressControlScreen.tsx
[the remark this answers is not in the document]
ivan-zynesis: I didn't quite followed the multiple addresses support. Is it an artificial limit we set for how many addresses to support. `discover()` will never end for large max length. Hmm, easiest will be storing addresses as persistence but the backup/restore will became different story.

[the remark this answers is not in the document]
fuxingloh: Should we set an artificial limit? @thedoublejay what are your thoughts on it?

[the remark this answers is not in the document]
thedoublejay: I tested it until 10 and i think it's too sluggish and long already. Probably let's just limit it to 4 first (which I think is too many already).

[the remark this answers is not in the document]
fuxingloh: Hmm, I think we can do 10 first? At least until we have a better wallet system.

mobile-app/app/screens/AppNavigator/screens/Balances/components/AddressControlScreen.tsx
fullstackninja864: After adding confirm pin, we already have loader saying `It may take a few seconds to securely encrypt your wallet...` this loader will be there until we don't discover all active address.

mobile-app/app/screens/AppNavigator/screens/Balances/components/AddressControlScreen.tsx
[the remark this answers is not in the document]
thedoublejay: @fullstackninja864 let's go with 10 first

mobile-app/app/screens/AppNavigator/screens/Balances/components/AddressControlScreen.tsx
```diff
@@ -0,0 +1,204 @@
...
+  )
+}
+
+export function AddressControlCard ({ onClose }: { onClose: () => void }): JSX.Element {
+  const { address, addressLength, setIndex, wallet } = useWalletContext()
+  const [availableAddresses, setAvailableAddresses] = useState<string[]>([])
+  const [canCreateAddress, setCanCreateAddress] = useState<boolean>(false)
+  const logger = useLogger()
+
+  const getAddresses = async (): Promise<void> => {
```
thedoublejay, line 72: We should rename this as we're not "getting" any addresses

thedoublejay (changes requested):

mobile-app/app/screens/WalletNavigator/screens/CreateWallet/PinConfirmation.tsx
```diff
@@ -47,12 +52,35 @@ export function PinConfirmation ({ route }: Props): JSX.Element {
...
+          if (type === 'restore') {
+            await discoverWalletAddresses(encrypted)
+          }
           await setWallet(encrypted)
         })
         .catch(logger.error)
     }, 50) // allow UI render the spinner before async task
   }
 
+  async function discoverWalletAddresses (data: WalletPersistenceDataI<EncryptedProviderData>): Promise<void> {
```
fuxingloh, line 64: Don't create another instance of Wallet, after `setWallet(encrypted)` on [line 58](https://github.com/DeFiCh/wallet/pull/1274/files#diff-0c3a3beb288ddbce90bb62f5e179694e7431c6feed2bda726bef9f5430f75339R58), use that newly initialized wallet and use it for discovery.

mobile-app/app/screens/AppNavigator/screens/Balances/components/AddressControlScreen.tsx
```diff
@@ -0,0 +1,204 @@
...
+    const addresses: string[] = []
+    for (let i = 0; i <= addressLength; i++) {
+      const account = wallet.get(i)
+      const address = await account.getAddress()
+      addresses.push(address)
+    }
+    setAvailableAddresses(addresses)
+    // incremented 1 to check if next account in the wallet is usable.
+    const length = addressLength + 1
```
fuxingloh, line 81:
```suggestion
    const next = addressLength + 1
```

fuxingloh (changes requested):

mobile-app/app/screens/WalletNavigator/screens/CreateWallet/PinConfirmation.tsx
fullstackninja864: @fuxingloh , in set wallet  we set the length of the wallet, here after that we directly switch the component (check the below logic)
```
if (wallets.length === 0) {
    return <WalletNavigator />
  }

  return (
    <WalletNodeProvider data={wallets[0]}>
      <WalletContextProvider api={WalletAddressIndexPersistence}>
        <PrivacyLock />
        <BottomSheetModalProvider>
          <TransactionAuthorization />
          <AppNavigator />
        </BottomSheetModalProvider>
      </WalletContextProvider>
    </WalletNodeProvider>
  )
  ```

In this switching there is now way we can find out this switching happens because of restore wallet or loading existing wallet. i.e. if we want to do some address discover in WalletNavigator but couldn't reach  WalletProvider. 
so by creating instance there once is less expensive operation so that we are discovering the active address and setting the count of active address before initialising the wallet instance we are going to use further.
 
cc @ivan-zynesis

thedoublejay (approved):

PR #1315, opened 2021-11-17
`loans` - add collateral

thedoublejay:

/kind feature

Very early iteration of add collateral

Fixes #

#### Developer Checklist:

- [ ] Unit tests*
- [ ] Added e2e tests*
- [ ] Added translations*


mobile-app/app/components/BottomSheetTokenList.tsx
```diff
@@ -30,79 +33,31 @@ export interface BottomSheetToken {
...
   return (
     <BottomSheetFlatList
-      data={tokenList}
-      renderItem={({ item }): JSX.Element => (
+      data={collateralTokens}
+      renderItem={({ item }: { item: CollateralItem }): JSX.Element => (
         <ThemedTouchableOpacity
+          disabled={new BigNumber(item.available).lte(0)}
           onPress={() => {
+            /*
```
kyleleow, line 49: I added this condition because composite swap has a use case that `onPress` it will not navigate but set the state of the parent

kyleleow (approved):

mobile-app/app/components/BottomSheetTokenList.tsx
thedoublejay: Ok noted. Will update this on a separate PR.

PR #1477, opened 2021-11-29
Fix bottom sheet on web 

### LINKED ISSUE #1476: Add/Edit Collateral Bottom Sheet not working in web
#### What happened:
The bottom sheet currently works well with mobile, but since we also have a non-mobile (chrome extension, web), it should also work as expected. 

- Bottom sheet should work on web

When navigating to: Loans -> Add Token as Collateral -> Select Token to Add -> Select any token -> the web version should still display without crashing

pierregee:

/kind fix

- This is to allow the bottom sheet to render on non-mobile devices while keeping the ios keyboard to push the UI up when opening

Fixes #1476

#### Additional comments?:
- There isn’t any web support on the library level as of now: https://github.com/gorhom/react-native-bottom-sheet/issues/11

#### Developer Checklist:

- [ ] Unit tests*
- [ ] Added e2e tests*
- [ ] Added translations*


mobile-app/app/components/WalletTextInput.tsx
```diff
@@ -20,12 +29,12 @@ interface IWalletTextInputProps {
   displayFocusStyle?: boolean
   containerStyle?: string
   onBlur?: () => void
+  hasBottomSheet?: boolean
```
pierregee, line 32: Tried to look into the library to detect if there's a "wrapping" bottom sheet, but it isn't exposed anywhere

PR #1485, opened 2021-11-29
update conversion to 2.1 DFI for vault fees

thedoublejay:

/kind fix

Fixes #

#### Developer Checklist:

- [ ] Unit tests*
- [ ] Added e2e tests*
- [ ] Added translations*


fuxingloh (commented):
If we use 3, users with more than 2 but less 3 will be affected. 

It's an anomaly but for this situation, it's also an anomaly for users with <2.1 DFI UTXO. Maybe just go with 2.01 instead?

pierregee (approved):

mobile-app/app/screens/AppNavigator/screens/Loans/screens/CreateVaultScreen.tsx
```diff
@@ -41,7 +41,7 @@ export function CreateVaultScreen ({
   const hasPendingBroadcastJob = useSelector((state: RootState) => hasBroadcastQueued(state.ocean))
   const DFIUtxo = useSelector((state: RootState) => DFIUtxoSelector(state.wallet))
   const DFIToken = useSelector((state: RootState) => DFITokenSelector(state.wallet))
-  const isConversionRequired = new BigNumber(2).gt(DFIUtxo.amount)
+  const isConversionRequired = new BigNumber(2.1).gt(DFIUtxo.amount)
```
pierregee, line 44: Noticed that it's 2.1 instead of 2.01, just confirming if it's intentional?

PR #1507, opened 2021-11-30
`announcement` - clickable link

### LINKED ISSUE #1502: `announcements` - clickable links
For our announcements banners, make it "clickable" when there's a URL present on the announcement item.  Add a caret when the banner is clickable
#### Why is this needed:
If we put a message that, "an update is needed", it should be able to redirect to app stores.

kyleleow:

/kind feature

- To allow banner to navigate to external url, e.g.: app store link for latest version, webpage for guide

Fixes #1502

#### Additional comments?:
- Change to get version from `package.json`, we should use version from `package.json` moving forward, since `nativeApplicationVersion` is always `null` for `web`

#### Developer Checklist:

- [ ] Unit tests*
- [ ] Added e2e tests*
- [ ] Added translations*


babel.config.js
```diff
@@ -18,7 +18,8 @@ module.exports = function (api) {
           '@screens': './mobile-app/app/screens',
           '@store': './shared/store',
           '@translations': './shared/translations',
-          '@tailwind': './mobile-app/app/tailwind'
+          '@tailwind': './mobile-app/app/tailwind',
+          '@package': './package.json'
```
thedoublejay, line 22: need to remove this

package.json
```diff
@@ -1,5 +1,6 @@
 {
   "name": "@defichain/wallet",
+  "version": "0.17.0",
```
thedoublejay, line 3: same here

mobile-app/app/screens/AppNavigator/screens/Balances/components/Announcements.tsx
```diff
@@ -15,33 +18,79 @@ export function Announcements (): JSX.Element {
...
+        testID='announcements_banner'
+        style={tailwind('px-4 py-3 flex-row items-center')} light={tailwind('bg-warning-50')}
+        dark={tailwind('bg-darkwarning-50')}
+        onPress={async () => await openURL(announcement.url)}
+      >
+        <ThemedIcon style={tailwind('mr-2')} iconType='MaterialIcons' name='campaign' size={22} />
+        <ThemedText style={tailwind('text-xs flex-auto')} testID='announcements_text'>
+          {`${announcement.content} `}
+          <View>
+            <ThemedIcon
```
thedoublejay, line 40: Can we make this fixed on the right side? Not at the end of the content but as a fixed button on the right (if link is existing)

thedoublejay (changes requested):

mobile-app/app/screens/AppNavigator/screens/Balances/components/Announcements.tsx
```diff
@@ -15,33 +18,79 @@ export function Announcements (): JSX.Element {
...
+            />
+          </View>
+        </ThemedText>
+      </ThemedTouchableOpacity>
+    )
   }
-  return <></>
+
+  return (
+    <ThemedView
```
thedoublejay, line 55: Probably we can make it as same component, to avoid having duplicate per content. Just disable the touchable opacity if there's no link

thedoublejay (changes requested):

babel.config.js
kyleleow: updated

package.json
kyleleow: updated

mobile-app/app/screens/AppNavigator/screens/Balances/components/Announcements.tsx
kyleleow: moved to right and increased to size `24`, but still pending final design from jonathan

mobile-app/app/screens/AppNavigator/screens/Balances/components/Announcements.tsx
kyleleow: updated to just use `<ThemedTouchableOpacity>`

shared/types/website.ts
```diff
@@ -13,6 +13,13 @@ export interface AnnouncementData {
    * Versioned matching represented as semver satisfies
    */
   version: string
+  url: {
```
thedoublejay, line 16: Probably better to make this optional

website/src/pages/api/v0/announcements.ts
```diff
@@ -16,14 +16,28 @@ export default async function handle (req: NextApiRequest, res: NextApiResponse<
       'zh-Hans': '请更新至最新版本应用程式使用去中心化贷款。',
       'zh-Hant': '請更新至最新版本應用程式使用去中心化貸款。'
     },
-    version: '<0.17.0'
+    version: '<0.17.0',
```
thedoublejay, line 19: Can you revert this change? Even with older users, they can't still update because code for new banner is not yet existing

website/src/pages/api/v0/announcements.ts
```diff
@@ -16,14 +16,28 @@ export default async function handle (req: NextApiRequest, res: NextApiResponse<
...
+    }
   }, {
     lang: {
       en: 'Decentralized loan is now available on the Light Wallet app.',
       de: 'Dezentrale Darlehen sind jetzt in der Light Wallet App verfügbar.',
       'zh-Hans': '去中心化贷款已在轻钱包 Light Wallet 正式推出',
       'zh-Hant': '去中心化貸款已在輕錢包 Light Wallet 正式推出'
     },
-    version: '>=0.17.0'
+    version: '>=0.17.0',
```
thedoublejay, line 34: Can be removed once optional is set

thedoublejay (changes requested):

website/src/pages/api/v0/announcements.ts
kyleleow: then instead of removing it entirely, why not we target after the version we added announcement, until the one version before the latest version? 

I check that 0.12 is when we added announcement banner

website/src/pages/api/v0/announcements.ts
kyleleow: updated

shared/types/website.ts
kyleleow: updated

thedoublejay (approved):

PR #1603, opened 2021-12-07
Convert Batch Cards to Flat List

### LINKED ISSUE #1593: `auctions` - update to flatlist
We need to optimize the display on main auctions page. We can first update to flatlist and create a new array to combine all batches into a single list so we don't loop twice.

pierregee:

/kind refactor

- Flattens the list before it is used in the component
- Purposely used one selector to minimize loops

Fixes #1593

#### Developer Checklist:

- [ ] Unit tests*
- [ ] Added e2e tests*
- [ ] Added translations*


thedoublejay (changes requested):
Need to add bottom padding on the main scrolling list

thedoublejay (approved):

PR #1729, opened 2021-12-14
Added scroll in quick bid screen for small screen

### LINKED ISSUE #1630: Auction: placing bid not possible on DE app version of iOS on iphone SE 2016 because scrolling does not work
#### What happened:
Placing a bid for auction is not possible, neither quick bid nor normal bid. This is due to the fact that the confirmation button can not be reached because scrolling does not work.

Scrolling is working and the button to place the bid is reachable. 

Using an iPhone SE 2016 model (small screen), this happens with german language. 
With English, quick bid scrolling does not work either.  Normal bid does work, because the button can be seen half. 
The general problem is that scrolling does not work somehow.

iOS app version 0.19 with enabled beta functions
image0
image1

fullstackninja864:

/kind fix

Fixes #1630

#### Developer Checklist:

- [x] Unit tests*
- [ ] Added e2e tests*
- [ ] Added translations*


pierregee (changes requested):
Is it possible to move the scrollbar a little bit so it's not overlapping in android?
image

pierregee (approved):

PR #1799, opened 2021-12-23
`e2e` - vault details

### LINKED ISSUE #1790: `e2e` - loans - vault details page

thedoublejay:

/kind chore

Fixes #1790

#### Developer Checklist:

- [ ] Unit tests*
- [ ] Added e2e tests*
- [ ] Added translations*


mobile-app/cypress/integration/functional/wallet/loans/vault_detail.spec.ts
```diff
@@ -0,0 +1,137 @@
+context('Wallet - Loans - Vault Details', () => {
+  let vaultId = ''
+
+  before(function () {
```
kyleleow, line 4: Wanted to discuss the usage of `beforeEach` vs `before`.

If we go with `before`, we couldn't pin point the exact test that is failing if the test case has dependency on any of the previous test.

E.g.: maybe `should check empty state` failed, and `should add loan` has dependency on it, but because the latter has dependency, it will fail too

kyleleow: But taking a step back, even if we decided to use `beforeEach` and ensure each test has no dependency with each other, we still wouldn't approve a PR at least ONE test failed.

This is more to test debugging purposes, at the cost of longer e2e run time(?), and more test code(?)

thedoublejay: My setup would be more on an e2e flow. To test the entire integration testing that's why I preferred using `before`. Since it's an entire "test scenario" I put it into one context. If it's not, I usually put it on another context. One advantage of this is there will be less bloat of code due to similar flow. Regarding your point about debugging, I don't have a hard time debugging it since you can actually watch the video on cypress dashboard.

With this approach, it reduces e2e time, reduces test code also. But I get your point about re-structuring. I'll see how we can re-structure it differently so we can easily pinpoint failing test cases.

mobile-app/cypress/integration/functional/wallet/loans/vault_detail.spec.ts
```diff
@@ -0,0 +1,137 @@
...
+  })
+
+  it('should check empty state', function () {
+    cy.getByTestID('vault_card_0').click()
+    cy.getByTestID('collateral_tab_COLLATERALS').click()
+    cy.getByTestID('button_add_collateral').should('exist')
+    cy.getByTestID('collateral_tab_LOANS').should('have.attr', 'aria-disabled')
+    cy.go('back')
+    cy.wait(2000)
+    cy.getByTestID('vault_card_0_edit_collaterals_button').click()
```
kyleleow, line 23: Since we are targeting vault details screen, why not we trigger add collateral using `button_add_collateral`?
Also because `vault_card_0_edit_collaterals_button` is already tested in `collateral.spec.ts`

thedoublejay: Oh right, i mislook that. Will update

mobile-app/cypress/integration/functional/wallet/loans/vault_detail.spec.ts
```diff
@@ -0,0 +1,137 @@
...
+    cy.getByTestID('vault_card_0').click()
+    cy.getByTestID('vault_id_section_min_ratio').contains('175%')
+    cy.getByTestID('text_vault_interest').contains('3')
+    cy.getByTestID('collateral_tab_DETAILS').click()
+    cy.getByTestID('text_min_col_ratio').contains('175.00')
+    cy.getByTestID('text_vault_interest_ratio').contains('3.00')
+  })
+})
+
+context('Wallet - Loans - Close Vault', () => {
```
kyleleow, line 76: can we add test to validate the access to edit loan scheme and close vault action based on different vault status?

kyleleow:
Also i feel like we can move these test (line 225 - 251) for vault details from `loan.spec.ts` to your new `vault_detail.spec.ts`
https://github.com/DeFiCh/wallet/blob/8c13a5cc478506589ad176c3c2b6b6af5ba5ddb7/mobile-app/cypress/integration/functional/wallet/loans/loans.spec.ts#L225

mobile-app/cypress/integration/functional/wallet/loans/vault_detail.spec.ts
kyleleow: Also test for various different UI in collateral, loan, and details tab in vault detail screen when the vault status changes

thedoublejay: It will be part of my liquidation test. But I'm having trouble setting up the parameters yet haha

kyleleow: ah ok

kyleleow (approved):

Issue #1917, opened 2022-01-14
`Loans` - Upon address switch, relevant vault not reflected


izzycsy:
https://user-images.githubusercontent.com/65014479/149468497-5e501b1f-d0d9-43d7-ab55-bc07e235cf37.mov

#### What happened:
Version: 0.22.0
• Upon switching address, the relevant vault is not reflected immediately but displays the previous selected address's vault data. 
• Correct data is reflected after: 
>  ~20secs of selecting another tab — not immediately
> leaving it to load for ~30secs

#### What you expected to happen:
• After switching address, respective address's vault should reflect accordingly

#### How to reproduce it (as minimally and precisely as possible):
• Start with address A > `Loans` > check vault ID > switch to address B > `Loans` > check vault ID (it reflects address A's vault)

#### Anything else we need to know?:
• Happens on both Android & iOS
Related: #1915

PR #1940, opened 2022-01-19
Fix vaults and poolpair update on switching address

### LINKED ISSUE #1915: "Your vaults" and "Your pool pairs" not updated after switching addresses
#### What happened: 
I have multiple addresses to manage different liquidity mining pools and vaults separately on the mobile wallet on Android (v0.22.0). Switching between the addresses leads to data not being displayed correctly in the areas "Your vaults" and "Your pool pairs". The displayed vault or pool pairs that belong to the previously selcted address are still shown instead. As a workaround you can select "swap" and immediately go back in the area "Your pool pairs" or select "Browse loan tokens" and immediately go back in the area "Your vaults". Or you can just restart the app. 

When switching addresses, all relevant screens should display only data relevant for the selected address. 

Create a second address, create a vault or add token to a lm pool pair and then switch addresses. 
 
Just a question: How can I as an experienced Java developer (20+ years) find a way into DeFiChain coding ;-)

pierregee:

/kind fix

- To refetch the`Your pool pair` data and `vaults` on wallet change based on the new address

Fixes #1915, #1917

#### Additional comments?:
- Since the fix isn't very straightfoward in poolpair, I'll explain it here. The issue with `Your Pool pair state` is with the flow of data updates  ```useEffect + RTK updates + current search string```
When wallet address change ->`useEffect` reruns `handleFilter` to get `filteredYourLPTokens` by search string WITHOUT waiting for the updated `yourLPTokens`. The fix is to use the `yourLPTokens` as the dependency array, but it causes multiple rerenders, so the logic is moved to its child component

#### Developer Checklist:

- [ ] Unit tests*
- [ ] Added e2e tests*
- [ ] Added translations*


thedoublejay (changes requested):
Can you add it also on
1. ManageBids
2. BorrowLoanTokenScreen

thedoublejay (approved):

PR #1957, opened 2022-01-21
chore(workflow): add semantic check on title

thedoublejay:

/kind chore

Added a workflow to validate PR titles based on [conventional commits guidelines](https://www.conventionalcommits.org/en/v1.0.0/). Using this [GH actions](https://github.com/amannn/action-semantic-pull-request) to run this workflow. 

For title, we will be using a custom one to not use a short form like `feat` and `perf`. We will use clearer terms so it's easier to understand (feature and performance) especially for new developers.

Lower case is enforced for subjects. Only PR title is checked since commits will be squashed. In the event where there's only one commit, this workflow will run on that single commit.

Important thing to note, it will only be run via GH actions and not on commit level (locally) so other open source developers can easily commit via online tools (Online VSC). 

Fixes #

#### Developer Checklist:

- [ ] Unit tests*
- [ ] Added e2e tests*
- [ ] Added translations*

### DIFF
--- a/.github/semantic.yml
+++ b/.github/semantic.yml
@@ -1,0 +1,1 @@
+titleOnly: true
--- a/.github/semantic.yml
+++ b/.github/semantic.yml
@@ -1,1 +1,8 @@
+name: Semantic Check
+
+on:
+  workflow_dispatch:
+  pull_request:
+    branches: [ main ]
+
 titleOnly: true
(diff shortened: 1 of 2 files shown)


kyleleow (commented):
I think you can fill up the description the PR with more details and reason why you want to do this and choice of `feature`

Also good to include the reference to the github action repo for future enhancement maybe

kyleleow (approved):

PR #1980, opened 2022-01-24
ci(translations): add workflow to output missing translations

### LINKED ISSUE #1973: `workflow` - missing language translations
On every PR build, we would like to trigger a CI build that:

1. Finds all English labels
2. Compares it with other languages
3. Reports the missing entries for each language

Related:
https://github.com/DeFiCh/wallet/pull/81
https://github.com/DeFiCh/wallet/issues/8
#### Why is this needed:
This will display any missing translations from the json files

kyleleow:

/kind feature

- To use of github workflow to detect missing translations, by using `ts-node` in one of the steps to run a `.ts` script to compare `de.json` with all other translation `.json` and report in github comment and slack
- If no missing translation is detected, PR comment will still be updated, however, slack message will not be sent

Fixes #1973

#### Developer Checklist:

- [ ] Unit tests*
- [ ] Added e2e tests*
- [ ] Added translations*

### DIFF
--- a/.github/workflows/translations-reporter.yml
+++ b/.github/workflows/translations-reporter.yml
@@ -38,7 +38,8 @@ jobs:
             ```json
             ${{ steps.translations.outputs.content }}
             ```
-      - name: Send missing translations to Slack workflow          
+      - name: Send missing translations to Slack workflow  
+        if: ${{ steps.translations.outputs.content.totalMissingCount != 0 }}
         uses: slackapi/slack-github-action@8157a0f4d70c5099e42dcff1258b19bdaabb030b
         id: slack_missing_translation_workflow
         with:
--- a/.github/workflows/translations-reporter.yml
+++ b/.github/workflows/translations-reporter.yml
@@ -39,7 +39,7 @@ jobs:
             ${{ steps.translations.outputs.content }}
             ```
       - name: Send missing translations to Slack workflow  
-        if: ${{ steps.translations.outputs.content.totalMissingCount != 0 }}
+        if: ${{ steps.translations.outputs.content['totalMissingCount'] != 0 }}
         uses: slackapi/slack-github-action@8157a0f4d70c5099e42dcff1258b19bdaabb030b
         id: slack_missing_translation_workflow
         with:
--- a/.github/workflows/translations-reporter.yml
+++ b/.github/workflows/translations-reporter.yml
@@ -38,17 +38,8 @@ jobs:
             ```json
             ${{ steps.translations.outputs.content }}
             ```
-      - id: set_missing_translation
-        run: |
-          content=`cat ./missing_translations.json`
-          # the following lines are only required for multi line json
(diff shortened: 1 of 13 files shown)


.github/workflows/translations-reporter.yml
```diff
@@ -0,0 +1,36 @@
+name: "Translations"
+
+on:
+  workflow_dispatch:
+  pull_request:
+    branches: [ main ]
```
thedoublejay, line 6: can you add this? so it will cancel ongoing runs in case a new one has been triggered - https://github.com/DeFiCh/wallet/blob/824158da6be57e0e9cc0697f2c1fefe5fe2c49e9/.github/workflows/pr-lint.yml#L9

.github/workflows/translations-reporter.yml
```diff
@@ -0,0 +1,36 @@
...
+      - run: npm ci
+      - name: Find missing translations
+        id: find_string
+        run: npm run translation:missing
+      - name: Read missing_translations.json
+        id: translations
+        uses: juliangruber/read-file-action@ebfa650188272343fef925480eb4d18c5d49b925
+        with:
+          path: ./missing_translations.json
+      - uses: marocchino/sticky-pull-request-comment@a8ba05e2dbda1247005adac58e1585b0177dc595
```
thedoublejay, line 27: Replace with @39c5b5dc7717447d0cba270cd115037d32d28443 (latest version)

shared/translations/reporter/index.ts
```diff
@@ -0,0 +1,90 @@
...
+function decodeLabel (label: string): string {
+  return Buffer.from(label, 'base64').toString('utf8')
+}
+
+function checkTranslations (baseTranslation: Map<string, string[]>, missingTranslations: MissingLanguage): MissingLanguage {
+  const localeToCheck = ['zh-Hans', 'zh-Hant', 'fr']
+  const languages = getAppLanguages().map(
+    language => language.locale
+  ).filter(
+    locale => localeToCheck.includes(locale)
```
thedoublejay, line 54: You want to filter the opposite? So we can just add new languages without modifying `localeToCheck`

thedoublejay (changes requested):
Can you run it also in the app if it doesn't break any translations feature? Since we have modified `initI18n`

.github/workflows/translations-reporter.yml
kyleleow: Updated

.github/workflows/translations-reporter.yml
kyleleow: Oh i am using the sha of the latest commit, will use the one from the latest tag instead. Updated

shared/translations/reporter/index.ts
kyleleow: Updated

thedoublejay (approved):

PR #2088, opened 2022-02-02
refactor(hooks): token price optimisation

### LINKED ISSUE #2032: optimize `TokenPrice.ts` to prevent slowness on most pages
Currently, we will be using this method to get DEX prices for most pages. It's currently too expensive and causes slowness on some pages (e.g, Auctions). We would like to optimize it to improve performance:

Some improvements that we can do (but not limited to):

- [x] Add useCallback
- [x] Add a cache object that will use block as keys. E.g, ( { 10000: { 'DFI-USDT' : 1000, 'DFI': 200 }

fullstackninja864:

/kind refactor

Fixes #2032

#### Developer Checklist:

- [ ] Unit tests*
- [ ] Added e2e tests*
- [ ] Added translations*


mobile-app/app/screens/AppNavigator/screens/Balances/hooks/TokenPrice.ts
```diff
@@ -50,19 +58,24 @@ export function useTokenPrice (): DexTokenPrice {
       const usdTokenB = getTokenPrice(pair.data.tokenB.symbol, tokenBAmount.toFixed(8))
       return usdTokenA.plus(usdTokenB)
     }
+    const key = `WALLET.${network}.TOKEN_PRICE_${symbol}`
```
thedoublejay, line 61: we can add block here

package.json
```diff
@@ -68,6 +68,7 @@
     "i18n-js": "^3.8.0",
     "install": "^0.13.0",
     "lodash": "^4.17.21",
+    "lru-cache": "^6.0.0",
```
thedoublejay, line 71: can we try this package?  https://www.npmjs.com/package/react-native-cache

This cache is meant for mobile usage

thedoublejay (changes requested):

kyleleow (dismissed):

mobile-app/app/api/cache/index.ts
```diff
@@ -0,0 +1,34 @@
+import Cache from 'lru-cache'
+
+class LruCache {
+  private readonly cache
+
+  constructor () {
+    // setting maxAge to 10 min
+    this.cache = new Cache({
```
thedoublejay, line 8: Add some options:
1. Max = to ensure we don't populate memory especially for older devices

mobile-app/app/screens/AppNavigator/screens/Balances/hooks/TokenPrice.ts
```diff
@@ -50,19 +58,24 @@ export function useTokenPrice (): DexTokenPrice {
...
+      return new BigNumber(result).multipliedBy(amount)
+    }
     // active price for walletTokens based on USDT
     const arbitraryPoolPair = getArbitraryPoolPair(symbol, 'USDT')
 
-    if (symbol === 'USDT') {
-      return new BigNumber(amount)
-    } else if (arbitraryPoolPair.length > 0) {
-      const { estimated } = calculatePriceRates(symbol, arbitraryPoolPair, amount)
+    if (arbitraryPoolPair.length > 0) {
```
thedoublejay, line 69:
```suggestion
    if (arbitraryPoolPair?.length > 0) {
```

thedoublejay (changes requested):
Still testing, will add other comments if there are any

mobile-app/app/screens/AppNavigator/screens/Balances/hooks/TokenPrice.ts
fullstackninja864: ` const arbitraryPoolPair = getArbitraryPoolPair(symbol, 'USDT')` this will always return array in return so its not required to use `arbitraryPoolPair?.length`

mobile-app/app/api/cache/index.ts
fullstackninja864: setting max to `max: 10000,`

mobile-app/app/screens/AppNavigator/screens/Balances/hooks/TokenPrice.ts
```diff
@@ -50,19 +58,24 @@ export function useTokenPrice (): DexTokenPrice {
...
+      const { aToBPrice, estimated } = calculatePriceRates(symbol, arbitraryPoolPair, amount)
+      // store price for each unit in cache
+      CacheApi.set(key, aToBPrice.toFixed(8))
```
pierregee, line 72: Just want to clarify if the intent here it to keep cache for 10mins since that is the default `maxAge`? If the tokenReserve change drastically before that, then the price rates will use the cached one. Thus, we will be displaying inaccurate rates for 10mins

pierregee: Oops. sorry I missed the key has the blockCount in it

pierregee: Maybe we can reduce the default maxAge since a new block is minted every 30seconds so there's no reason to keep the `expired` ones in the device's memory for 10mins? Do you think 2 minutes is more reasonable as `maxAge`

fullstackninja864: @pierregee sometimes our blockchain takes 5-6 min to block a new block(based on historic data), But to lower memory usage we will set it to 5

PR #2203, opened 2022-02-17
feature(ui-ux): auctions improvements

thedoublejay:

/kind chore

Improves performance of Auctions page

Currently, we call `fetchPoolPairs` on `useTokenPrice` hook. This results into multiple API fetching and re-rendering when this hook is called. To solve this, we will now call `fetchPoolPairs` on the StatsProvider so it's only called once. The rest will just rely on the selector to get updated values.

Second change, is to load only 5 auctions at start. Once end is reached, it will add 5 more. This is to minimize rendering and improve performance on cold starts

Fixes #

#### Developer Checklist:

- [ ] Unit tests*
- [ ] Added e2e tests*
- [ ] Added translations*

### DIFF
--- a/shared/contexts/StatsProvider.tsx
+++ b/shared/contexts/StatsProvider.tsx
@@ -6,8 +6,11 @@ import { block } from '@store/block'
 import { useNetworkContext } from '@shared-contexts/NetworkContext'
 import { useWhaleApiClient } from './WhaleContext'
 import { useLogger } from '@shared-contexts/NativeLoggingProvider'
+import { fetchPoolPairs } from '@store/wallet'
 
 export function StatsProvider (props: PropsWithChildren<any>): JSX.Element | null {
+  const blockCount = useSelector((state: RootState) => state.block.count)
+  const client = useWhaleApiClient()
   const { network } = useNetworkContext()
   const logger = useLogger()
   const isPolling = useSelector((state: RootState) => state.block.isPolling)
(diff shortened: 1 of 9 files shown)


pierregee (approved):

shared/contexts/StatsProvider.tsx
```diff
@@ -54,6 +64,11 @@ export function StatsProvider (props: PropsWithChildren<any>): JSX.Element | nul
     }
   }, [api, interval, network, dispatch])
 
+  // Global polling based on blockCount and network, so no need to fetch per page
+  useEffect(() => {
```
pierregee, line 68: 👌

kyleleow, line 68: No comment on the functionality-wise, this branch is too fast that I thought i was using a samsung galaxy s32 ultra pro max 4k phone

But i have one comment on adding the pooling for `fetchPoolPairs` here because it doesn't contribute to the main purpose of `StatsProvider`, unless we change the file name to something more generic, would it work the same if you were to move this to maybe `<Main>`?

thedoublejay: Oh yeah good suggestion, let me move it. After this release, I'll move `fetchTokens` also. But it will affect lots of pages so not this release

thedoublejay: Updated

kyleleow (approved):

PR #2300, opened 2022-03-01
fix(ui-ux): cancelling a search will reset button filter (#2273)

### LINKED ISSUE #2273: `DEX` - cancelling a search will reset button filter
#### What happened:
Cancelling a search after typing at least one character would reset the button filter and includes all pool pairs

Cancelling a search would reset the list based on the current active filter

- DEX screen, tap DUSD pairs filter
- Tap search icon from top left
- Enter 'a'
- Tap `X` to cancel search
- Noticed the pool pairs displayed under DUSD pairs filter includes all pool pairs

Related to #2271

rox1338:

/kind fix

Fixes the issue described in #2273 by applying the button filter after cancelling the search.
Also I removed the usage of debounce with a delay in handleFilter as it doesn't seem beneficial and makes problem in combination with handleButtonFilter which doesn't use debounce with a delay.

Fixes #2273

#### Developer Checklist:

- [ ] Unit tests*
- [ ] Added e2e tests*
- [ ] Added translations*


thedoublejay:
Hi @rox1338, eslint is failing because `debounce` that was removed is still on the file. Can you update your PR to remove this? Thank you!

mobile-app/app/screens/AppNavigator/screens/Dex/DexScreen.tsx
```diff
@@ -118,8 +117,7 @@ export function DexScreen (): JSX.Element {
   const [filteredAvailablePairs, setFilteredAvailablePairs] =
     useState<Array<DexItem<PoolPairData>>>(pairs)
   const [isSearching, setIsSearching] = useState(false)
-  const handleFilter = useCallback(
-    debounce((searchString: string) => {
```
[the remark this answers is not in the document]
rox1338: I did some more research on debounce and agree with you.
I'll try to come up with a different solution.

kyleleow:
Hello @rox1338, thanks for opening this PR.
After checking the changes, they look good to me. However, now there is a slight UX inconsistency in terms of the poolpairs being displayed to the user during the following steps:

1. user tap `DUSD pairs` button filter
2. list shows DUSD pairs
3. user then tap the search icon, but without typing any thing in the search input
4. the list that was shown now are still DUSD pairs <--- i think here we should show all pairs because the search is for all pairs
5. user type any input to search, list is filtered accordingly without being filtered from button selection <-- this is correct
6. user then clear the search input string, list shown now is all unfiltered pairs <-- this is correct

On step 4, the list shown is not in consistent as step 6, but in both steps the search input is empty, could you help to address this well?

rox1338:
> Hello @rox1338, thanks for opening this PR. After checking the changes, they look good to me. However, now there is a slight UX inconsistency in terms of the poolpairs being displayed to the user during the following steps:
> 
> 1. user tap `DUSD pairs` button filter
> 2. list shows DUSD pairs
> 3. user then tap the search icon, but without typing any thing in the search input
> 4. the list that was shown now are still DUSD pairs <--- i think here we should show all pairs because the search is for all pairs
> 5. user type any input to search, list is filtered accordingly without being filtered from button selection <-- this is correct
> 6. user then clear the search input string, list shown now is all unfiltered pairs <-- this is correct
> 
> On step 4, the list shown is not in consistent as step 6, but in both steps the search input is empty, could you help to address this well?

I actually thought about that and decided to implement it the way I did it because I thought it would be weird that the result list changes just by pressing the search button.
But I'm open for implementing it the way you mentioned.
Do you have an example how this is solved in other applications? (Best practices)

kyleleow:
@rox1338 let's display an empty list when search input is active and when it has no input string yet. This is to avoid causing any confusion to the user and to decouple from whatever filter that was being selected on.

A similar behaviour can be seen from other apps like twitter and reddit.

rox1338:
> @rox1338 let's display an empty list when search input is active and when it has no input string yet. This is to avoid causing any confusion to the user and to decouple from whatever filter that was being selected on.
> 
> A similar behaviour can be seen from other apps like twitter and reddit.

I agree.
Done.

kyleleow:
@rox1338 looking good! Can you also help to apply the same flow to `Your pool pairs`?
Right now if user is viewing `Your pool pairs` tab, when tap on the search icon, it would still show all of the user's pool pairs

rox1338:
> @rox1338 looking good! Can you also help to apply the same flow when `Your pool pairs` is the active tab?

Did it.
I would suggest to refactor these files. As you can see the logic for filtering 'your pool pairs' is in PoolPairCards.tsx and the logic for filtering the available pool pairs is in DexScreen.tsx.

mobile-app/app/screens/AppNavigator/screens/Dex/DexScreen.tsx
```diff
@@ -180,18 +184,21 @@ export function DexScreen (): JSX.Element {
...
+      setIsSearching(true)
+      handleFilter(searchString)
+    }
   }, [searchString, hasFetchedPoolpairData])
 
   // Update local state - filter available pair when pairs update
   useEffect(() => {
-    if (searchString !== undefined && searchString.trim().length > 0) {
-      handleFilter(searchString)
+    if (showSearchInput) {
```
kyleleow, line 195: If possible we would try to avoid nested if-condition statement, this is mainly for readability 
e.g.: we can do an early return for `if (!showSearchInput) {... return}`

mobile-app/app/screens/AppNavigator/screens/Dex/components/PoolPairCards/PoolPairCards.tsx
```diff
@@ -116,17 +116,29 @@ export function PoolPairCards ({
...
     setIsSearching(false)
-    setFilteredYourPairs(
-      yourPairs.filter((pair) =>
-        pair.data.displaySymbol
-          .toLowerCase()
-          .includes(debouncedSearchTerm.trim().toLowerCase())
-      ).sort((a, b) =>
-        availablePairs.findIndex(x => x.data.id === a.data.id) -
-        availablePairs.findIndex(x => x.data.id === b.data.id
+    if (showSearchInput) {
```
kyleleow, line 119: same here, can we do an early return to avoid nested if-condition?

mobile-app/app/screens/AppNavigator/screens/Dex/components/PoolPairCards/PoolPairCards.tsx
```diff
@@ -116,17 +116,29 @@ export function PoolPairCards ({
...
+      if (debouncedSearchTerm !== undefined && debouncedSearchTerm.trim().length > 0) {
+        setFilteredYourPairs(
+          yourPairs.filter((pair) =>
+            pair.data.displaySymbol
+              .toLowerCase()
+              .includes(debouncedSearchTerm.trim().toLowerCase())
+          ).sort((a, b) =>
+            availablePairs.findIndex(x => x.data.id === a.data.id) -
```
kyleleow, line 127: aware that the filtering logic is repeated in `DexScreen`, but as for this PR sake, we could refactor the sorting logic into a function in the mean time

kyleleow (changes requested):
Thanks for following up, added a few comments. 
As for refactoring the repeated logic, we are more than happy if you wish to contribute in the same PR!
Otherwise, you could file an issue and we would address it in the future

mobile-app/app/screens/AppNavigator/screens/Dex/DexScreen.tsx
rox1338: Not sure whether that makes it more readable, but I will do that

mobile-app/app/screens/AppNavigator/screens/Dex/components/PoolPairCards/PoolPairCards.tsx
rox1338: yes

mobile-app/app/screens/AppNavigator/screens/Dex/components/PoolPairCards/PoolPairCards.tsx
rox1338: ok

rox1338:
> Thanks for following up, added a few comments. As for refactoring the repeated logic, we are more than happy if you wish to contribute in the same PR! Otherwise, you could file an issue and we would address it in the future

I would be happy to do the refactoring but would like to finish this bugfix first and do the refactoring in a follow-up issue.
I created an issue for that #2340

kyleleow:
@rox1338 looks good, I'll change the target branch to merge your changes into a separate branch in order to trigger our configured workflow, before merging into `main`
Thanks again for your contribution 🥳

kyleleow (approved):

PR #2303, opened 2022-03-02
feature(ui-ux): `collaterals` - add resulting vault share when adding/removing collateral

### LINKED ISSUE #1867: `collaterals` - add resulting vault share when adding/removing collateral
When adding/removing collaterals, display resulting vault share.
#### Why is this needed:
For each collateral added, it would be helpful for users to understand how it will affect their current vault. This will also prevent some issues, like getting locked collateral due to improper balance.

chloezxyy:

/kind feature

This PR will add an extra section called 'Vault %' to display the resulting vault share, so that users will understand how their input affects their current vault.

BottomSheetModal of 'Select tokens to add' and 'How many tokens to add' size has been increased slightly both in web and mobile view so that it's easier to view

Fixes #1867

#### Developer Checklist:

- [ ] Unit tests*
- [x] Added e2e tests*
- [x] Added translations*


mobile-app/app/components/BottomSheetTokenList.tsx
```diff
@@ -89,7 +93,8 @@ export const BottomSheetTokenList = ({
                     onButtonPress: navigateToScreen.onButtonPress,
                     collateralFactor: new BigNumber(item.factor ?? 0).times(100),
                     isAdd: true,
-                    vault
+                    vault,
+                    ...(isCollateralItem(item) && { collateralItem: item })
```
thedoublejay, line 97: Why are you spreading the return of this?

mobile-app/app/screens/AppNavigator/screens/Loans/components/AddOrRemoveCollateralForm.tsx
```diff
@@ -114,6 +113,17 @@ export const AddOrRemoveCollateralForm = memo(({ route }: Props): JSX.Element =>
     setCollateralValue(amount)
   }
 
+  const currentBalance = vault?.collateralAmounts?.find((c) => c.id === token.id)?.amount ?? '0'
```
thedoublejay, line 116: Are we missing the initial scenario here? Like, if it's the only VS, we display 100% immediately. The result might be the same. I just want to save on calculation in case it's 100%.

thedoublejay: Meaning, we can short-circuit by not needing to calculate it in case there are no other collaterals.

mobile-app/app/screens/AppNavigator/screens/Loans/components/AddOrRemoveCollateralForm.tsx
```diff
@@ -242,23 +255,51 @@ export const AddOrRemoveCollateralForm = memo(({ route }: Props): JSX.Element =>
...
-        <ThemedText
-          style={tailwind('font-semibold')}
+          style={tailwind('font-semibold pr-2')}
           light={hasInvalidColRatio ? tailwind('text-gray-300') : colors.light}
           dark={hasInvalidColRatio ? tailwind('text-gray-300') : colors.dark}
           testID='resulting_collateralization'
         >{hasInvalidColRatio ? translate('components/AddOrRemoveCollateralForm', 'N/A') : `${resultingColRatio.toFixed(2)}%`}
         </ThemedText>
-      </ScrollView>
+      </View>
```
thedoublejay, line 302: Why we removed scrollview?

thedoublejay (changes requested):
Another thing to note, since we're calculating it here. Should we just pass the vault share from here to the confirm page? So in the confirm page, we don't need to calculate it again since they should display the same data.

mobile-app/app/components/BottomSheetTokenList.tsx
pierregee: Suggested to use this to @chloezxyy to be able to pass the collateralItem to ensure it won't be `undefined`, but it needs to check if the type is `CollateralItem` via ts typeguard so it won't affect other components using the BottomSheet that doesn't have CollateralItem e.g composite swap

mobile-app/app/screens/AppNavigator/screens/Loans/components/AddOrRemoveCollateralForm.tsx
chloezxyy: It's not removed, its shifted to line 291, where the vault % will be the last item that can be seen on the BottomSheetModal without scrolling

thedoublejay (changes requested):
@chloezxyy 
1. Set default value on initial load
2. Test all scenarios
3. Can you try to vertical center the icon and vault share percentage? Looks a bit off now

thedoublejay (approved):

Issue #2225, opened 2022-02-21
`POC` - Local database


thedoublejay:
#### What would you like to be added:
One technical limitation we have currently is the lack of centralized database due to the nature of our product (decentralized). Without it, we can't do any features that requires syncing with a database. To solve this, we would like to have some sort of local database. 

However, this local database will not be synced across different devices. But, we can have an import/export feature that will be used for this user preferences file.

Need to be checked if this is technical feasible.

PR #2304, opened 2022-03-02
feature(core): create local storage

thedoublejay:

/kind feature

1. Allows the use of a local storage json file for native devices and localStorage for web.
2. Uses `expo-file-system` to create a local file for storage purposes
3. User preferences will be called on change of network or wallets
4. Ability to fetch setUserPreferences via dispatch
5. setAddresses can be done to overwrite the preferences file
6. useAppDispatch hook to be able to fully use thunks, should be chainable like a promise

Fixes https://github.com/DeFiCh/wallet/issues/2225

#### Developer Checklist:

- [ ] Unit tests*
- [ ] Added e2e tests*
- [ ] Added translations*


mobile-app/app/api/local_storage/provider/provider.native.ts
```diff
@@ -0,0 +1,54 @@
...
+    console.log(e)
+  }
+}
+
+async function getUserPreferences (network: EnvironmentNetwork): Promise<UserPreferences> {
+  let userPreferences = null
+  if (FileSystem.documentDirectory != null) {
+    const directory = getFileName(network, FileSystem.documentDirectory)
+    await ensureFileExist(directory)
+    const fileContents = await FileSystem.readAsStringAsync(directory)
```
kyleleow, line 33: If line 32 threw an error, line 33 will still be reached, shall we move the try-catch from `ensureFileExist` to `getUserPreferences` ?

mobile-app/app/api/local_storage/provider/provider.native.ts
```diff
@@ -0,0 +1,54 @@
...
+
+async function setUserPreferences (network: EnvironmentNetwork, userPreferences: UserPreferences): Promise<void> {
+  try {
+    if (FileSystem.documentDirectory !== null) {
+      const directory = getFileName(network, FileSystem.documentDirectory)
+      await ensureFileExist(directory)
+      await FileSystem.writeAsStringAsync(directory, JSON.stringify(userPreferences), { encoding: 'utf8' })
+    }
+  } catch (e) {
+    console.log(e)
```
kyleleow, line 47: Should we use the centralized `Logging` api?

shared/store/userPreferences.ts
```diff
@@ -0,0 +1,55 @@
...
+    preferences
+  }: { network: EnvironmentNetwork, preferences: UserPreferences }) => {
+    await LocalStorageProvider.setUserPreferences(network, preferences)
+  }
+)
+
+export const setAddresses = createAsyncThunk(
+  'userPreferences/setAddresses',
+  async (addresses: LabeledAddress) => {
+    return addresses
```
kyleleow, line 37: Since we are not performing any async task here, think we don't need `createAsyncThunk`. Can we move the logic from `extraReducers` to `reducer`?

kyleleow (changes requested):
Can you help to add some unit tests to `mobile-app/app/api/secured/index.test.ts`? Like how we did for `SecuredStoreAPI`

shared/store/userPreferences.ts
thedoublejay: I moved it here because I wanted to make it an async thunk, meaning, after you set the address via dispatch, you will then call the setUserPrefences (to store it in the local storage). I'm not sure if we move it to extra reducer, it will work. But we can test it out once there's a UI side to integrate (set address flow).

mobile-app/app/api/local_storage/provider/provider.native.ts
thedoublejay: Oh yeah will update!

mobile-app/app/api/local_storage/provider/provider.native.ts
thedoublejay: Hmmm yeah good point, let me see what I can do

thedoublejay: updated!

thedoublejay:
> Can you help to add some unit tests to `mobile-app/app/api/secured/index.test.ts`? Like how we did for `SecuredStoreAPI`

Will add this test on another PR so integration won't be blocked

PR #2348, opened 2022-03-08
feature(ui-ux): portfolio value breakdown

### LINKED ISSUE #2280: `balances` - total portfolio value and rename to `portfolio`
We would like to include all locked amounts (e.g, tokens used as vault collaterals) on the Total Portfolio Value. Currently, only Token amounts (from your balances) are included on the USD value. This will help also on monitoring the health of their loans.

- [x] Rename Menu bar and Title to Portfolio. Since it's not going to be your "Balances" only but all your portfolio including those locked.
image
image

- [x] Add collateral values in Total Portfolio Value
- [x] Add a breakdown of available and locked in vault that will be visible via expand
image

fullstackninja864:

/kind feature

This PR includes all locked amounts (e.g, tokens used as vault collaterals) on the Total Portfolio Value. Currently, only Token amounts (from your balances) are included on the USD value. This will help also on monitoring the health of their loans.

Fixes #2280

#### Developer Checklist:

- [x] Unit tests*
- [x] Added e2e tests*
- [x] Added translations*


thedoublejay (changes requested):
@fullstackninja864 initial feedbacks:
1. Address switching will change the calculation of portfolio. Some delay on showing full amount.
2. Label should be locked in vault(s)
3. Locked in vault amount is wrong.

thedoublejay (changes requested):
Page is updated everytime, so snapshot is dynamically updated failing unit tests.

mobile-app/app/screens/AppNavigator/screens/Balances/BalancesScreen.tsx
```diff
@@ -57,57 +60,77 @@ export function BalancesScreen ({ navigation }: Props): JSX.Element {
   }, [height, wallets])
 
   useEffect(() => {
+    fetchPortfolioData()
```
thedoublejay, line 63: Should we also have isFocused here?

thedoublejay (changes requested):

mobile-app/app/screens/AppNavigator/screens/Balances/BalancesScreen.tsx
fullstackninja864: we fetching token data on continuous basis so that we will have latest balance details on all screens, so here we are not adding is focused flag here, added note in the code

thedoublejay (approved):

PR #2363, opened 2022-03-09
fix(core): change payback loan's max to available amount if balance is not enough

### LINKED ISSUE #2351: Loan payback - Change MAX value of NON-DUSD loans
#### What happened:
- Issue: When paying back loan, the max button for Non-DUSD Loans is always the outstanding loan balance

- The max value for DUSD Loans should always be the outstanding balance
- The max value for Non-DUSD Loans should be the :
1. Outstanding balance if available balance is enough,
2. Max available balance if available balance is not enough

Take note: The design for this will be changed in the next iterations.

Linked - https://github.com/DeFiCh/wallet/issues/2167

pierregee:

/kind fix

The max value for Non-DUSD Loans should be the :
- Outstanding balance if available balance is enough,
- Max available balance if available balance is not enough

Fixes #2351

#### Developer Checklist:

- [ ] Unit tests*
- [ ] Added e2e tests*
- [x] Added translations*

### DIFF
--- a/mobile-app/app/screens/AppNavigator/screens/Loans/screens/PaybackLoanScreen.tsx
+++ b/mobile-app/app/screens/AppNavigator/screens/Loans/screens/PaybackLoanScreen.tsx
@@ -79,9 +79,10 @@ export function PaybackLoanScreen ({
   const canUseOperations = useLoanOperations(vault?.state)
   const client = useWhaleApiClient()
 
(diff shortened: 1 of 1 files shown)


thedoublejay (changes requested):
Still not working, default outstanding balance is max even though balance is not enough for non DUSD loans

thedoublejay (approved):

PR #2398, opened 2022-03-15
feature(core): remove preselect on to token

### LINKED ISSUE #2346: `token detail` - swap option
Previously, in the Token Detail page, we only have swap with DFI/DUSD due to being able to use direct paths only. With the addition of composite swap, we can allow to pre-select the "FROM" token based on the token selected, then display all tokens that can be swapped in the "TO" list.

1. Portfolio -> Click on a token
- [ ] If there's a swap option (e.g, Non LP tokens), currently it is Swap with DFI/Swap with DUSD. Replace it with Swap token
- [ ] It should be redirected to pool swap page, and FROM token prefilled with the selected
- [ ] FROM token should be disabled

pierregee:

/kind feature

- Make `TO` token enabled WITHOUT preselected token when navigating from Token Detail
- Keep the `FROM` token disabled

Fixes #2346

#### Developer Checklist:

- [ ] Unit tests*
- [x] Added e2e tests*
- [x] Added translations*


mobile-app/app/screens/AppNavigator/screens/Dex/CompositeSwap/CompositeSwapScreen.tsx
```diff
@@ -221,32 +222,48 @@ export function CompositeSwapScreen ({ route }: Props): JSX.Element {
...
-        tokenId: pair.data.tokenB.id,
-        available: new BigNumber(pair.data.tokenB.reserve),
-        token: {
-          displaySymbol: pair.data.tokenB.displaySymbol,
-          symbol: pair.data.tokenB.symbol,
-          name: '' // not available in API
-        },
-        reserve: pair.data.tokenB.reserve
-      }, 'TO')
+      if (tokenSelectOption.from.isPreselected) {
```
kyleleow, line 241: Can we do an early return here at line 240? To avoid nested if-block.
e.g.: `if (pair !== undefined) return`

pierregee: Agree. Upddated

thedoublejay (approved):

kyleleow (approved):

thedoublejay (changes requested):
Failing one e2e test case

thedoublejay (approved):

PR #2434, opened 2022-03-18
refactor(core): integrate swappable tokens from whale

### LINKED ISSUE #2433: DEX - composite swap - swappable tokens
- There should be no changes in functionality.
- Light wallet is filtering the `From` and `To` tokens on the client side. This change will use the logic on the whale side
image

#### Why is this needed:
- To improve performance of composite swap - fetching FROM and TO Tokens

pierregee:

/kind refactor

- To Tokens needs to be checked if there's a path from `fromToken` to `toToken` which light wallet is determining in the client side manually through poolpair hops. This PR will move its logic to the whale side

Fixes #2433

#### Developer Checklist:

- [ ] Unit tests*
- [ ] Added e2e tests*
- [ ] Added translations*


mobile-app/app/screens/AppNavigator/screens/Dex/hook/SwappableTokens.ts
```diff
@@ -0,0 +1,95 @@
...
+        return swappableFromTokens
+    }, [poolpairs])
+
+    useEffect(() => {
+        if (fromTokenId !== undefined) {
+            dispatch(fetchSwappableTokens({
+                client, fromTokenId: fromTokenId === '0_unified' ? '0' : fromTokenId
+            }))
+        }
+    }, [blockCount, fromTokenId])
```
fullstackninja864, line 89: Can we implement cashing mechanism here to avoid repetitive api calls when user reselect the token?
for eg: If user first select BTC as from token then change it to ETH and then again resect BTC (no new block get mined in this switching).
In this case we can save one api call.
cc @thedoublejay

fullstackninja864 (changes requested):

mobile-app/app/screens/AppNavigator/screens/Dex/hook/SwappableTokens.ts
```diff
@@ -0,0 +1,95 @@
...
+                    },
+                    reserve: token.reserve
+                }
+            }).sort((a, b) => b.available.minus(a.available).toNumber())
+
+        return swappableFromTokens
+    }, [poolpairs])
+
+    useEffect(() => {
+        if (fromTokenId !== undefined) {
```
fullstackninja864, line 84: Also can we use `isFocused` here to avoid re-renders

mobile-app/app/screens/AppNavigator/screens/Dex/hook/SwappableTokens.ts
thedoublejay: Yep, we can cache this definitely. By symbol/block

thedoublejay: Actually, I think we don't need to call this on every block update. Swappable tokens rarely updates (usually when there are only new tokens).

thedoublejay (changes requested):
Added some comments on thread with Harsh

mobile-app/app/screens/AppNavigator/screens/Dex/hook/SwappableTokens.ts
pierregee: Updated. used `useFocusEffect` and replaced `useMemo` as it runs even if screen ain't focused

mobile-app/app/screens/AppNavigator/screens/Dex/hook/SwappableTokens.ts
pierregee: Updated. Removed blockCount on `dep array` to reduce rerenders, but kept the blockCount in the caching

mobile-app/app/screens/AppNavigator/screens/Dex/hook/SwappableTokens.ts
```diff
@@ -0,0 +1,95 @@
...
+import { useDispatch, useSelector } from 'react-redux'
+import BigNumber from 'bignumber.js'
+import { RootState } from '@store'
+import { DexItem, fetchSwappableTokens, tokensSelector } from '@store/wallet'
+import { useWhaleApiClient } from '@shared-contexts/WhaleContext'
+import { BottomSheetToken } from '@components/BottomSheetTokenList'
+import { TokenState } from '../CompositeSwap/CompositeSwapScreen'
+
+interface TokenPrice {
+    toTokens: BottomSheetToken[]
```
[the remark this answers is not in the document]
pierregee: Updated!

fullstackninja864 (approved):

thedoublejay (approved):

PR #2556, opened 2022-04-05
feature(core): add edit & delete in address book

### LINKED ISSUE #2523: `address book` - edit existing labels
Linked: https://github.com/DeFiCh/wallet/issues/2240

Ability to edit both labels and addresses on the address book. This must also use the same validations like before (e.g, valid address, unique etc.). Editing will need to prompt passcode also.

### LINKED ISSUE #2524: `address book` - delete address entries
Linked: https://github.com/DeFiCh/wallet/issues/2523
Linked: https://github.com/DeFiCh/wallet/issues/2240

Ability to delete existing address entries. Take note, this will also need a passcode prompt when doing this operation.

kyleleow:

/kind feature

- To allow edit and delete from address book

Fixes #2523
Fixes #2524 

#### Developer Checklist:

- [ ] Unit tests*
- [x] Added e2e tests*
- [x] Added translations*

### DIFF
--- a/mobile-app/app/screens/AppNavigator/screens/Balances/screens/AddOrEditAddressBookScreen.tsx
+++ b/mobile-app/app/screens/AppNavigator/screens/Balances/screens/AddOrEditAddressBookScreen.tsx
@@ -11,7 +11,6 @@ import { useNetworkContext } from '@shared-contexts/NetworkContext'
 import { useWalletNodeContext } from '@shared-contexts/WalletNodeProvider'
 import { RootState } from '@store'
 import { authentication, Authentication } from '@store/authentication'
-import { removeFromAddressBook } from '@store/userPreferences'
 import { tailwind } from '@tailwind'
 import { translate } from '@translations'
 import { useCallback, useEffect, useLayoutEffect, useState } from 'react'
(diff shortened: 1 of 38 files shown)


mobile-app/app/screens/AppNavigator/screens/Balances/screens/AddOrEditAddressBookScreen.tsx
```diff
@@ -23,10 +24,12 @@ export function AddOrEditAddressBookScreen ({ route, navigation }: Props): JSX.E
   const {
     title,
     onSaveButtonPress,
-    addressLabel
+    address,
+    addressLabel,
+    isAddNew
   } = route.params
   const [labelInput, setLabelInput] = useState(addressLabel?.label)
```
pierregee, line 31: image
No issues found, but to fix the warning, it can't be `undefined` on the initial state,  `addressLabel?.label ?? ''`
If it'll break tests and take time, I think it's fine to keep the warning for now and just create a ticket

kyleleow: Actually saw the warning but only on web, not on mobile. I also didnt fallback to empty string because there is a validation on clean state, and dirty state with empty string. The ideal solution is probably to use the controlled input like `react-hook-form` in Send screen

thedoublejay (approved):

PR #2625, opened 2022-04-12
feature(ui-ux): add timer hook for execution countdown

kyleleow:

/kind feature

- To add hook to display day-hour-minute countdown
- To add conditions to block continue and confirm button when blockCount equals to or exceeds execution block
- Display `To be confirmed` as resulting balance and estimated to receive
- To update confirm swap screen to display relevant info when it is future swap

Fixes #

#### Developer Checklist:

- [ ] Unit tests*
- [ ] Added e2e tests*
- [ ] Added translations*


mobile-app/app/screens/AppNavigator/screens/Dex/CompositeSwap/ConfirmCompositeSwapScreen.tsx
```diff
@@ -100,6 +103,22 @@ export function ConfirmCompositeSwapScreen ({ route }: Props): JSX.Element {
...
+      }
+    } else {
+      if (!isFutureSwap) {
+        return translate('screens/ConfirmCompositeSwapScreen', 'Swap')
+      } else {
+        return translate('screens/ConfirmCompositeSwapScreen', 'Future swap')
+      }
+    }
+  }
+
```
pierregee, line 121: Might not have big impact, but let's still apply fair early approach since most of the time people will use this for normal swap without conversion. 

Should we remove the **Convert and future swap** since there's no `DFI` in future swap?


```suggestion
  function getTransactionType(): string {
    if (conversion?.isConversionRequired !== true) {
      if (!isFutureSwap) {
        return translate('screens/ConfirmCompositeSwapScreen', 'Swap')
      } else {
        return translate('screens/ConfirmCompositeSwapScreen', 'Future swap')
      }
    } else {
      return translate('screens/ConfirmCompositeSwapScreen', 'Convert & swap')
    }
  }
  
```

```

pierregee (changes requested):

mobile-app/app/screens/AppNavigator/screens/Dex/CompositeSwap/ConfirmCompositeSwapScreen.tsx
kyleleow: yep make sense, will update

kyleleow: updated

pierregee (approved):

PR #2662, opened 2022-04-19
fix(ui-ux): `balances` align token amount text on the right

chloezxyy:

/kind fix

Fixes #

#### Developer Checklist:

- [ ] Unit tests*
- [ ] Added e2e tests*
- [ ] Added translations*


thedoublejay:
Let's include this next release

chloezxyy:
Token amount should be aligned to the right even if token value text is longer than token amount text

image

pierregee (changes requested):
LGTM, just need to update the unit test

pierregee (approved):

PR #2671, opened 2022-04-20
feature(ui-ux): add address book row ui in settings page

chloezxyy:

/kind feature

- Rename section title to NETWORK to GENERAL 
- Network row and Address book row will be under GENERAL
- Create a row for Address book item in Settings page 
- Address rows should not be clickable when navigating from the Settings page

Fixes #

#### Additional comments?:
- Will add e2e when all UI is ready

#### Developer Checklist:

- [ ] Unit tests*
- [ ] Added e2e tests*
- [ ] Added translations*


mobile-app/app/screens/AppNavigator/screens/Balances/screens/AddressBookScreen.tsx
```diff
@@ -34,6 +34,9 @@ export function AddressBookScreen ({ route, navigation }: Props): JSX.Element {
   const addressBook = userPreferences.addressBook
   const [isEditing, setIsEditing] = useState(false)
 
+  // disable address selection touchableopacity from settings page
+  const disableOpacity = !!((selectedAddress === undefined && onAddressSelect === undefined))
```
kyleleow, line 38: - does it work if we remove double bracket and double `!!`?
- can we rename variable to `disableAddressSelect`?

mobile-app/app/screens/AppNavigator/screens/Balances/screens/AddressBookScreen.tsx
```diff
@@ -46,7 +49,10 @@ export function AddressBookScreen ({ route, navigation }: Props): JSX.Element {
     if (hasPendingJob || hasPendingBroadcastJob) {
       return
     }
-    onAddressSelect(address)
+    // condition to make address component unclickable from settings page
+    if (address !== undefined && onAddressSelect !== undefined) {
```
kyleleow, line 53: can we reuse the variable you created at the beginning?

mobile-app/app/screens/AppNavigator/screens/Balances/screens/AddressBookScreen.tsx
chloezxyy: works when removing one bracket, `!!` must stay 
yes of course we can rename it :>

mobile-app/app/screens/AppNavigator/screens/Balances/screens/AddressBookScreen.tsx
chloezxyy: can't reuse since the logic is checking for different conditions

PR #2684, opened 2022-04-21
remove collateral from list if active price is 0 #2623

### LINKED ISSUE #2623: `collaterals` - remove collateral from list if active price is 0
#### What happened:
During hard forks, we sometimes add collaterals on the list of collateral tokens. Sometimes, when the "price" from the oracle is not ready, users won't be able to add until the price from oracle is live.

To solve this issue, we would like to remove the collateral token on the list of collaterals IF the active price is 0. 

To make it easy, you can develop it on Playground network. You may go to Settings => Network => Playground => Click on Playground again. There will be options there to click/top up tokens. With this, you will be able to create a vault.

The easy way is to just filter the list of collateral tokens if active price is 0.

rox1338:
/kind fix

During hard forks, we sometimes add collaterals on the list of collateral tokens. Sometimes, when the "price" from the oracle is not ready, users won't be able to add until the price from oracle is live.

Fixes #2623

#### Additional comments?:
Is there some sort of automated test for the EditCollateralScreen?
Can we introduce a token with price 0 on playground?
I couldn't test the actual 0-check as there is no token with price 0. But I did a test with the check being for >1 to see whether the filtering works with that value. (And it did ;-) )


mobile-app/app/screens/AppNavigator/screens/Loans/screens/EditCollateralScreen.tsx
```diff
@@ -99,7 +99,8 @@ export function EditCollateralScreen ({
       ...c,
       available: getTokenAmount(c.token.id)
     }
-  }).sort((a, b) => b.available.minus(a.available).toNumber()))
+  }).filter((collateralItem) => new BigNumber(getActivePrice(collateralItem.token.symbol, collateralItem.activePrice)) > 0)
```
thedoublejay, line 102: I think BigNumber has native gt check. Maybe you can use that?

mobile-app/app/screens/AppNavigator/screens/Loans/screens/EditCollateralScreen.tsx
```diff
@@ -225,7 +226,7 @@ export function EditCollateralScreen ({
                   onCloseButtonPress: dismissModal,
                   navigateToScreen: {
                     screenName: 'AddOrRemoveCollateralForm',
-                    onButtonPress: onAddCollateral
+                    onButtonPress: onAddCollateral as any
```
thedoublejay, line 229: Why do we need this?

rox1338: because I get this error:
229:36  error    Promise-returning function provided to property where a void return was expected                                                    @typescript-eslint/no-misused-promises

PR #2698, opened 2022-04-25
feature(core): retry flags api on error

### LINKED ISSUE #2697: White screen stuck when flags API returns 404
#### What happened:
- If the flags API has an error, light wallet won't retry and will be stuck at white screen.

- The app should be kept on white screen (this is the current behaviour),

 - If the flags API starts working, it should automatically load the app without the need to close -> open the app)
 - If the flags API doesn't work, it will keep on retrying for approximately 30 seconds. After 30 seconds, if API is still not working, then will display the app with empty flags which means some of the features enabledthrough flags will be disabled

pierregee:

/kind fix

- Add retry every 10s when flags API returns an error

Fixes #2697

#### Developer Checklist:

- [ ] Unit tests*
- [ ] Added e2e tests*
- [ ] Added translations*


mobile-app/app/contexts/FeatureFlagContext.tsx
```diff
@@ -102,7 +108,7 @@ export function FeatureFlagProvider (props: React.PropsWithChildren<any>): JSX.E
 
   return (
     <FeatureFlagContext.Provider value={context}>
-      {props.children}
+      {!isError && !isLoading && props.children}
```
thedoublejay, line 111: can we add this condition on the return <></> ?

thedoublejay (changes requested):

mobile-app/app/contexts/FeatureFlagContext.tsx
```diff
@@ -39,6 +39,12 @@ export function FeatureFlagProvider (props: React.PropsWithChildren<any>): JSX.E
   const [enabledFeatures, setEnabledFeatures] = useState<FEATURE_FLAG_ID[]>([])
   const { network } = useNetworkContext()
 
+  if (isError) {
+    setTimeout(() => {
+      prefetchPage({})
+    }, 10000)
+  }
+
   if (!isError) {
```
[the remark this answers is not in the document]
pierregee: Updated!

mobile-app/app/contexts/FeatureFlagContext.tsx
pierregee: Updated!

thedoublejay (approved):

PR #2728, opened 2022-04-28
fix(ui-ux): display locked collateral in portfolio

### LINKED ISSUE #2702: `portfolio` - if token is used as 100% collateral. it's not displayed on tokens list
If the token is used 100% collateral. It's not displayed on tokens list on portfolio page. Currently, it's only displayed if there's an available balance.

e.g:
If you have 10 ETH, all of it is used as collateral, it will not be displayed on portfolio page. If you remove 1 ETH as collateral, then it will be displayed.

We need to display the token on the list even without any balance as long as it's locked as collateral.

kyleleow:

/kind fix

- To display locked collateral in portfolio regardless of percentage locked

Fixes #2702

#### Developer Checklist:

- [ ] Unit tests*
- [x] Added e2e tests*
- [ ] Added translations*


shared/store/wallet.ts
```diff
@@ -98,16 +95,6 @@ export const setTokenSymbol = (t: AddressToken): WalletToken => {
   }
 }
 
-const associateTokens = (tokens: TokenData[]): AssociatedToken => {
-  const result: AssociatedToken = {}
-  tokens.forEach(token => {
-    if (token.isDAT) {
```
pierregee, line 104: We are filtering out `isDAT` because the API returns multiple displaySymbol and some of it are not the DAT

Here's my 2 cents:
 - If this will be removed, then we need to update `tokenSelectorByDisplaySymbol` to only return `isDAT` or else it would return incorrect ids

- Also if this filtering will be removed on this level, we always have to be mindful when using `allTokens` to always use `isDAT` unless oceanAPI will update it on their end to always filter it out



image

kyleleow: thanks for the context

kyleleow: updated

pierregee (dismissed):

pierregee (dismissed):
Issues raised are resolved and tokens with only locked collateral is now displaying

mobile-app/app/screens/AppNavigator/screens/Balances/BalancesScreen.tsx
```diff
@@ -154,6 +154,37 @@ export function BalancesScreen ({ navigation }: Props): JSX.Element {
...
+      return dstTokens
+    }
+
+    const dstTokenSymbols = dstTokens.map(token => token.symbol)
+    lockedTokens.forEach((_lockedBalance, symbol) => {
+      const tokenExist = dstTokenSymbols.includes(symbol)
+      if (!tokenExist) {
+        const tokenData = allTokens.find(token => token.symbol === symbol)
+        if (tokenData !== undefined) {
+          dstTokens.push({
```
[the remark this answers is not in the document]
kyleleow: `Map` doesn't have `reduce()` method, but if you're talking about immutability, i have update to return a new array.

mobile-app/app/screens/AppNavigator/screens/Balances/hooks/TokenLockedBalance.ts
```diff
@@ -15,10 +15,10 @@ export interface LockedBalance {
 
 /**
  *
- * @param symbol optional token symbol
- * @returns Map of all token's locked balance or single object of symbol passed
+ * @param symbol optional token displaySymbol
+ * @returns Map of all token's locked balance or single object of displaySymbol passed
```
thedoublejay, line 19: I think there's a reason why it was kept to Symbol. Might be the API?

thedoublejay: Can you double check first? Just to ensure we don't cause any regressions

thedoublejay (changes requested):

mobile-app/app/screens/AppNavigator/screens/Balances/hooks/TokenLockedBalance.ts
kyleleow: we kept `symbol` last time because `getTokenPrice()` requires symbol.
If you look at line 41, I am still calling `getTokenPrice()` using `symbol`, that part is unchanged, i only changed the `displaySymbol` to be used as the key for the returned `Map`

I have also cross-checked in mainnet, the number checks out for all of usd, DFI and BTC value of tokens, LPs, and dTokens

thedoublejay (approved):

PR #2841, opened 2022-05-24
refactor(core): replaced use of activateAfterBlock in CollateralItem tokens with activePrice

fullstackninja864:

/kind refactor

Previously we used `activateAfterBlock` to identify the item is collateral token or normal token, in this PR we replaced the logic with use of `activePrice` and checking the item having `activePrice` obj, if yes then its collateral token otherwise its normal one.

Fixes #

#### Developer Checklist:

- [ ] Unit tests*
- [ ] Added e2e tests*
- [ ] Added translations*


thedoublejay (changes requested):
@kyleleow DUSD doesn't have activePrice right? So this PR will cause regressions

thedoublejay:
Should not be needed anymore with this - https://github.com/JellyfishSDK/jellyfish/pull/1481

PR #2879, opened 2022-05-30
chore(core): bump react-redux package

### LINKED ISSUE #2835: update to `react-redux@8.x.x`
Currently, we're using `react-redux@7.x.x`. Need to check compatibility if we can update to `react-redux@8.x.x`.

pierregee:

/kind chore

Upgrade to react-redux v8 

**React redux v8 upgrade key points:**
**Goal of react-redux v8** is to be compatible with React 18 (LW is currently using react v17.0.2)
- Since react-redux v8 will work with React 16.8 and higher, we can already upgrade react-redux v7 to v8 without the need to upgrade to React 18
- LW cannot update to React 18 yet because [RN (React Native) New Architecture](https://reactnative.dev/blog/2022/03/30/version-068#opting-in-to-the-new-architecture) isn't ready yet. For more information: https://www.youtube.com/watch?v=FZ0cG47msEk&t=1530s
- Followed the types from official docs for useAppDispatch lint issue https://redux-toolkit.js.org/tutorials/typescript, this update also requires to use the pre-typed custom useAppDispatch

**References and other related discussions/announcements:**
- https://reactjs.org/blog/2022/03/29/react-v18.html
- https://github.com/reduxjs/react-redux/issues/1740

The only issue found on this upgrade is the infinite loop on the Feature Provider because of max retries which is fixed by using `useEffect` hook

Fixes #2835

#### Developer Checklist:

- [ ] Unit tests*
- [ ] Added e2e tests*
- [ ] Added translations*


thedoublejay:
To add on React@18 update, we update react based on Expo support (and react-native in a way).

mobile-app/app/contexts/FeatureFlagContext.tsx
```diff
@@ -41,14 +41,16 @@ export function FeatureFlagProvider (props: React.PropsWithChildren<any>): JSX.E
   const { network } = useNetworkContext()
   const [retries, setRetries] = useState(0)
 
-  if (isError && retries < MAX_RETRY) {
-    setTimeout(() => {
+  useEffect(() => {
```
thedoublejay, line 44: Why was this moved to useEffect?

thedoublejay (changes requested):

mobile-app/app/contexts/FeatureFlagContext.tsx
pierregee: After the bump of react-redux, it starte throwing infinite loop error

thedoublejay (changes requested):
You can also remove `@types/react-redux` as it's already included on the main package

mobile-app/app/contexts/FeatureFlagContext.tsx
pierregee: For reference: 
image

PR #2920, opened 2022-06-02
loans - active loans values overflow in vault card #2847

### LINKED ISSUE #2847: `loans` - active loans values overflow in vault card
- To not overflow UI out of bound

#### Why is this needed:
- UI consistency

- Loans screen > Your vaults tab

image

rox1338:
/kind feature

Fixes #2847

### DIFF
--- a/mobile-app/app/components/TokenIconGroup.tsx
+++ b/mobile-app/app/components/TokenIconGroup.tsx
@@ -20,21 +20,14 @@ interface TokenIconGroupProps {
 
 export function TokenIconGroup (props: TokenIconGroupProps): JSX.Element {
   const additionalIcon = BigNumber.max(props.symbols?.length - props.maxIconToDisplay, 0)
-
-  function calculateRightOffset (): number {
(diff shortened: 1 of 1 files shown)


rox1338:
FYI @thedoublejay

mobile-app/app/components/TokenIconGroup.tsx
```diff
@@ -20,7 +20,12 @@ interface TokenIconGroupProps {
 
 export function TokenIconGroup (props: TokenIconGroupProps): JSX.Element {
   const additionalIcon = BigNumber.max(props.symbols?.length - props.maxIconToDisplay, 0)
-  const rightOffset = props.offsetContainer === true ? ((props.symbols.length - 1) * -5) - 1 : 0
+  let rightOffset
```
thedoublejay, line 23: can you set this to 0 by default? Then change it if props.offsetContainer is true

rox1338: sure

PR #2986, opened 2022-06-14
fix(store): filtered out v1 tokens and poolpairs 

### LINKED ISSUE #2977: hide all tokens from stock split
Because of stock splits, we will have tokens and pool pairs with `/v1`. Filter it out on LW and Scan.

Screenshot_20220607-160808_DeFiChain Wallet
Screenshot_20220610-161133_DeFiChain Wallet

siradji:

/kind fix

Fixes #2977 

#### Developer Checklist:

- [ ] Unit tests*
- [ ] Added e2e tests*
- [ ] Added translations*


shared/store/wallet.ts
```diff
@@ -112,7 +112,9 @@ export const fetchPoolPairs = createAsyncThunk(
   'wallet/fetchPoolPairs',
   async ({ size = 200, client }: { size?: number, client: WhaleApiClient }): Promise<DexItem[]> => {
     const pairs = await client.poolpairs.list(size)
-    return pairs.map(data => ({ type: 'available', data }))
+    return pairs
+      .filter(pair => !pair.displaySymbol.includes('/v1'))
```
thedoublejay, line 116: Use `symbol`. I'm afraid historical tokens might not have all attributes (that a normal token have). But symbol will always be there.

thedoublejay: Also can you add a comment here, just so anyone knows the reason behind it

shared/store/wallet.ts
```diff
@@ -128,7 +130,8 @@ export const fetchTokens = createAsyncThunk(
   'wallet/fetchTokens',
   async ({ size = 200, address, client }: { size?: number, address: string, client: WhaleApiClient }): Promise<{ tokens: AddressToken[], allTokens: TokenData[], utxoBalance: string }> => {
     const tokens = await client.address.listToken(address, size)
-    const allTokens = await client.tokens.list(size)
```
thedoublejay, line 131: Does it handle also possible swap tokens?

thedoublejay (changes requested):

shared/store/wallet.ts
siradji: yes i believe. I checked the swap tokens and dAmz/v1 is being filtered out

pierregee (changes requested):
image
I believe this also needs to filter out on the swap feature (?)

siradji:
> image
> I believe this also needs to filter out on the swap feature (?)

@pierregee it is done!

thedoublejay (approved):

kyleleow:
Want to also add a test for the swappable token?

kyleleow:
Created an [issue](https://github.com/DeFiCh/wallet/issues/3002) for future e2e enhancement, current PR looks good.

kyleleow (approved):

mobile-app/cypress/e2e/functional/wallet/dex/compositeswap.spec.ts
```diff
@@ -174,10 +174,14 @@ context('Wallet - DEX - Composite Swap with balance', () => {
     cy.getByTestID('token_select_button_FROM').click()
     cy.getByTestID('select_DFI_value').should('have.text', '20.00000000')
     cy.getByTestID('select_DFI').click().wait(1000)
-
     cy.getByTestID('token_select_button_TO').click()
     cy.getByTestID('select_dLTC').should('exist')
     cy.getByTestID('select_DFI').should('not.exist')
+    cy.findByTestId('bottom_sheet_token_list').within(() => {
```
kyleleow, line 180: functionality-wise it should make no difference, but just for code consistency-sake
```suggestion
    cy.getByTestID('bottom_sheet_token_list').within(() => {
```

kyleleow (approved):

PR #3016, opened 2022-06-21
feature(core): add custom service provider

### LINKED ISSUE #2753: ocean endpoint as user config
To give users the possibility to setup their own ocean and use this one instead of the official one, I'd love to see a config possibility to override the ocean base url. Maybe incl function test and automatic fallback to the official ocean.

#### Why is this needed:
As we're growing fast, more and more bots are used, the Ocean is hard at it's limit. To reduce the load on ocean and having a better user experice, this feature would be very handy.

pierregee:

/kind feature

- Allow changing of service provider on each environment

Fixes #2753 

#### Additional comments?:
- LW has an existing known behaviour wherein we keep the app at white screen unless the API responds after 3 retries. Since the user might be stuck forever when the custom endpoint is not working. 
**Workaround**: If it's custom endpoint, it will load the app regardless if API will respond or not (it'll display the loaders instead). This will allow the user to correct the custom endpoint

#### Developer Checklist:

- [x] Unit tests*
- [ ] Added e2e tests*
- [x] Added translations*


mobile-app/app/screens/AppNavigator/screens/Settings/components/ResetButton.tsx
```diff
@@ -0,0 +1,87 @@
...
+      additionalMessageUrl: defaultUrl
+    }
+    dispatch(authentication.actions.prompt(auth))
+  }, [dispatch, navigation])
+
+  // pop up box
+  const onPress = async (): Promise<void> => {
+    WalletAlert({
+      title: translate('screens/ServiceProviderScreen', 'Reset Service Provider'),
+      message: translate('screens/ServiceProviderScreen', 'In doing so, you will be reverted back to Light wallet\'s default endpoint. Would you like to continue?'),
```
thedoublejay, line 55: Instead of Light wallet, maybe "DeFiChain Wallet" is better?

mobile-app/cypress/e2e/functional/wallet/settings/settings.spec.ts
```diff
@@ -227,3 +229,167 @@ context('Wallet - Settings - Address Book', () => {
...
+    })
+
+    it('should submit valid custom service provider', () => {
+      cy.getByTestID('endpoint_url_input').clear().type(url.custom).wait(3000)
+      cy.getByTestID('button_submit').click().wait(1000)
+      cy.getByTestID('pin_authorize').type('000000').wait(5000)
+      cy.go('back')
+    })
+
+    // it('should display Custom on Server', () => {
```
thedoublejay, line 319: should we remove?

thedoublejay (changes requested):

shared/translations/languages/it.json
```diff
@@ -316,7 +316,8 @@
     "Sign to save address": "Firma per salvare l'indirizzo",
     "Sign to delete address": "Firma per cancellare l'indirizzo",
     "Address Book": "Rubrica",
-    "GENERAL": "GENERALE"
+    "GENERAL": "GENERALE",
+    "Server": ""
```
thedoublejay, line 320: empty

thedoublejay (changes requested):

mobile-app/cypress/e2e/functional/wallet/settings/settings.spec.ts
pierregee: Updated. Added a comment on why it's commented

pierregee: To follow fix for this issue ^ 
image

shared/translations/languages/it.json
pierregee: Updated!

mobile-app/cypress/e2e/functional/wallet/settings/settings.spec.ts
pierregee: Fixed!

PR #3026, opened 2022-06-22
feature(ui-ux): add new header and wallet stacks v2

kyleleow:

/kind feature

- To add new header title
- To add wallet stacks v2 that will uses the new header title

Fixes #

#### Developer Checklist:

- [ ] Unit tests*
- [ ] Added e2e tests*
- [ ] Added translations*


mobile-app/app/screens/Main.web.tsx
```diff
@@ -9,25 +9,28 @@ import { PlaygroundNavigator } from './PlaygroundNavigator/PlaygroundNavigator'
...
-      <View style={styles.phone}>
-        <RootNavigator />
-      </View>
-
-      {env.name !== EnvironmentName.Production && (
-        <View style={[styles.phone, tailwind('bg-white ml-2')]}>
-          <NavigationContainer theme={DeFiChainTheme}>
-            <PlaygroundNavigator />
-          </NavigationContainer>
+    <SafeAreaProvider>
```
thedoublejay, line 19: didn't we have this already somewhere?

thedoublejay (changes requested):

mobile-app/app/constants/Theme.ts
```diff
@@ -7,9 +7,8 @@ export function getDefaultTheme (isLight: boolean): Theme {
     ...defaultTheme,
     colors: {
       ...defaultTheme.colors,
-      primary: isLight ? '#ff00af' : '#EE2CB1',
```
thedoublejay, line 10: wouldn't this cause regression issues?

thedoublejay (changes requested):

mobile-app/app/screens/Main.web.tsx
kyleleow: We have it on `Main.tsx` but not on web

mobile-app/app/constants/Theme.ts
kyleleow: updated

thedoublejay (approved):

PR #3324, opened 2022-08-17
fix(ui-ux): potential fix for disappearing bottom sheet

pierregee:

/kind fix

- Trial and error - fix. After adding/removing props, I noticed that when adding a custom handle, the very intermittent error is fixed. 
- How to replicate the issue? Quick Bid or anything that uses the useBottomSheet hook open and close multiple times

Fixes #

#### Developer Checklist:

- [ ] Unit tests*
- [ ] Added e2e tests*
- [ ] Added translations*

### DIFF
--- a/mobile-app/app/components/BottomSheetWithNav.tsx
+++ b/mobile-app/app/components/BottomSheetWithNav.tsx
@@ -5,7 +5,7 @@ import {
   StackNavigationOptions,
   TransitionPresets
 } from '@react-navigation/stack'
-import { BottomSheetBackdropProps, BottomSheetBackgroundProps, BottomSheetModal } from '@gorhom/bottom-sheet'
+import { BottomSheetBackdropProps, BottomSheetBackgroundProps, BottomSheetHandleProps, BottomSheetModal } from '@gorhom/bottom-sheet'
 import { BottomSheetModalMethods } from '@gorhom/bottom-sheet/lib/typescript/types'
 import { NavigationContainer, Theme } from '@react-navigation/native'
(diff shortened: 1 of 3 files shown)


mobile-app/app/components/BottomSheetWithNav.tsx
```diff
@@ -75,7 +75,11 @@ export const BottomSheetWithNav = React.memo((props: BottomSheetWithNavProps): J
   )
 })
 
-export const BottomSheetWebWithNav = React.memo((props: BottomSheetWithNavProps & { isModalDisplayed: boolean, modalStyle?: { [other: string]: any} }): JSX.Element => {
+const EmptyHandleComponent = (_props: BottomSheetHandleProps): JSX.Element => {
```
thedoublejay, line 78: can we use props instead of having underscore

mobile-app/app/components/BottomSheetWithNavV2.tsx
```diff
@@ -67,6 +67,10 @@ export const BottomSheetWithNavV2 = React.memo((props: BottomSheetWithNavProps):
   )
 })
 
+const EmptyHandleComponent = (_props: BottomSheetHandleProps): JSX.Element => {
```
thedoublejay, line 70: same

thedoublejay (changes requested):

G9000:
LGTM. I tested it on one of my affected modals and its works like a charm 😎

mobile-app/app/components/BottomSheetWithNav.tsx
pierregee: removed, since it's unused

PR #3384, opened 2022-08-31
Remove screen name translations in navigators

### LINKED ISSUE #3383: Remove screen name translations
Many Screen names will be translated, which is not neccessary and makes it impossible to extend deep linking.
Example:
https://github.com/DeFiCh/wallet/blob/2578036cb3c47354eddcc5dd828fe5e4238de217/mobile-app/app/screens/AppNavigator/screens/Portfolio/PortfolioNavigator.tsx#L153
#### Why is this needed:
Deep linking needs unique keys per item and tree.

mbochmann:

/kind refactor

Names of Screens in Navigators should not be translated because they will be used as keys for deep linking.

Fixes #3383

#### Developer Checklist:


kyleleow:
Thanks for creating this PR, we have implemented the same changes in another PR #3691, which we are also resetting the navigation stack after language switch.

PR #3427, opened 2022-09-08
fix(ui-ux): swap v2 slippage tolerance fix

chloezxyy:

/kind fix

- Addition of logic to save custom slippage value in local storage 
- Disable continue button if no slippage value is selected or input value 

Fixes #

#### Developer Checklist:

- [ ] Unit tests*
- [ ] Added e2e tests*
- [ ] Added translations*


mobile-app/app/api/persistence/custom_slippage-tolerance_storage.ts
```diff
@@ -0,0 +1,18 @@
+import AsyncStorage from "@react-native-async-storage/async-storage";
+
+const KEY = "WALLET.CUSTOM_SLIPPAGE_TOLERANCE";
+
+async function set(
+  customSlippageTolerance: NonNullable<string>
+): Promise<void> {
+  await AsyncStorage.setItem(KEY, customSlippageTolerance);
+}
+async function get(): Promise<string> {
```
thedoublejay, line 10: can't we use the existing one?

thedoublejay (changes requested):

mobile-app/app/api/persistence/custom_slippage-tolerance_storage.ts
chloezxyy: This is to store the true/false value of whether a custom value is used or not, not the actual value of the slippage tolerance input value

thedoublejay: Yes but if I'm not mistaken, the old logic does it as well. If there's a value, then use custom. Otherwise, not. Would it work similarly?

chloezxyy: Will remove the custom storage and handle it on the front instead

pierregee (approved):

PR #3445, opened 2022-09-12
chore(e2e): update compositeswap.specs.ts

chloezxyy:

Fixes #

#### Developer Checklist:

- [ ] Unit tests*
- [ ] Added e2e tests*
- [ ] Added translations*


mobile-app/cypress/e2e/functional/wallet/dex/compositeswap.spec.ts
```diff
@@ -173,8 +176,29 @@ context("Wallet - DEX - Instant/Future Swap - tabs and dropdowns", () => {
     );
   });
 
+  // the value of price rates keeps changing - can't assert an absolute value
```
pierregee, line 179: Could you try with DFI -> dETH? So the difference is big?
image
 
 The test for the value is crucial or the price rate bug could recur in the future. If absolute value isn't possible, we can make use of `checkValueWithinRange`

pierregee (changes requested):

pierregee (approved):

PR #3686, opened 2022-10-19
feature(ui-ux): display only brand new pools in DEX

chloezxyy:
To only display brand new pool pair in DEX, existing pools with stock split will not be considered as a new pool and will not be displayed anymore

Fixes #

#### Developer Checklist:

- [ ] Unit tests*
- [ ] Added e2e tests*
- [ ] Added translations*


mobile-app/app/screens/AppNavigator/screens/Dex/DexScreen.tsx
```diff
@@ -302,7 +302,8 @@ export function DexScreen(): JSX.Element {
     Array<DexItem<PoolPairData>>
   >([]);
   useEffect(() => {
-    const sorted = pairs.map((item) => item).slice(-5);
+    // display brand new poolpairs, stocksplit stocks are considered as old poolpairs
+    const sorted = pairs.filter((x) => !x.hasStockSplit).slice(-5);
```
pierregee, line 306: can we use `p` or `pair` instead of x  😛

shared/store/wallet.ts
```diff
@@ -211,9 +213,27 @@ export const wallet = createSlice({
...
+        state.poolpairs = action.payload
+          .filter(({ data }) => !data.symbol.includes("/v1"))
+          .map((item) => {
+            return {
+              ...item,
+              // poolpairs with added hasStockSplit boolean indicator for post stocksplit
+              hasStockSplit: stockSplitPairs.some((pair) => {
+                return (
+                  pair.data.symbol === `${item.data.symbol}/v1` ||
+                  item.data.symbol.includes("/v1")
```
pierregee, line 232:
```suggestion
                  pair.data.symbol === `${item.data.symbol}/v1`
```
condition not needed right cos all the string with `v1` is filtered out on line 224?

pierregee (changes requested):

shared/store/wallet.ts
chloezxyy: If item.data.symbol.includes("/v1") is not included in the condition then the hasStockSplit will be set to false instead of true

pierregee (approved):

pierregee (approved):
LGTM. But it's more of a decision if it's ok to not consider pair that was "stocksplit" as a new poolpair @thedoublejay

PR #3691, opened 2022-10-20
feature(ui-ux): reset nav stack upon switching language

lykalabrada:
- Reset navigation stack when switching language
- KY: removing translation on base route such that navigating AFTER switching language without specifying the base route would not cause unexpected navigation

Fixes #

#### Additional comments?:
_Known issue_: Intermittently, after switching to a different language, user is redirected to either Settings screen or Portfolio screen. Ideally it should consistently redirect to just one screen.

#### Developer Checklist:

- [ ] Unit tests*
- [ ] Added e2e tests*
- [ ] Added translations*


mobile-app/app/screens/AppNavigator/BottomTabNavigator.tsx
```diff
@@ -147,3 +147,30 @@ export const AppLinking = {
     },
   },
 };
+
+export const bottomTabRoutes = [
```
kyleleow, line 151:
```suggestion
export const bottomTabDefaultRoutes = [
```

PR #3705, opened 2022-10-25
fix(ui-ux): trailing >2dp for NEW POOLS

veralygit:
This PR is a fix for >2dp for NEW POOLS after adding LP or perform a swap

Fixes #
1. update NumberFormat to include that value is a NumericString so that format maintains after pool value is updated

#### Additional comments?:
To reproduce 
Perform a swap on dex, while tx is processing, note: NEW POOLS section
Leave app running and check on NEW POOLS section if there is still > 2dp

#### Developer Checklist:

- [ ] Unit tests*
- [ ] Added e2e tests*
- [ ] Added translations*


mobile-app/app/screens/AppNavigator/screens/Dex/components/DexScrollable.tsx
```diff
@@ -156,6 +156,7 @@ function TotalLiquidityValue({
           decimalScale={0}
```
kyleleow, line 156: let's not use `decimalScale`, instead we control the decimal places using `BigNumber`.
At line 155, use this to display currency `getPrecisedCurrencyValue(value)`

veralygit: updated line 155 to :
`value={BigNumber(value).toFixed(2)}`
removed `decimalScale={0}`

mobile-app/app/screens/AppNavigator/screens/Dex/components/DexScrollable.tsx
```diff
@@ -152,10 +152,10 @@ function TotalLiquidityValue({
         </ThemedTextV2>
       ) : (
         <NumberFormat
-          value={value}
-          decimalScale={0}
+          value={BigNumber(value).toFixed(2)}
           thousandSeparator
           displayType="text"
+          valueIsNumericString
```
kyleleow, line 158: does it work if we remove this?

veralygit: if removed valueIsNumericString it works as well to 2dp
image

i also found out that when we use toFixed(2) it would append .00 to whole values. Would that be fine too 
image

PR #3712, opened 2022-10-25
fix(e2e): updated and fix  loans e2e

G9000:
- Include sorting test in the loans e2e 
- add liquidity e2e  fix

Fixes #

#### Developer Checklist:

- [ ] Unit tests*
- [x] Added e2e tests*
- [ ] Added translations*

### DIFF
--- a/mobile-app/cypress/e2e/functional/wallet/loans/loans.spec.ts
+++ b/mobile-app/cypress/e2e/functional/wallet/loans/loans.spec.ts
@@ -42,7 +42,9 @@ function checkTokensSortingOrder(
             .invoke("text")
             .then((text) => {
               const secondVal = text.split("% Interest")[0];
-              expect(Number(firstVal)).to.be.lessThan(Number(secondVal));
+              expect(Number.parseInt(firstVal)).to.be.lessThan(
+                Number.parseInt(secondVal)
+              );
             });
         });
       break;
@@ -55,7 +57,9 @@ function checkTokensSortingOrder(
             .invoke("text")
             .then((text) => {
               const secondVal = text.split("% Interest")[0];
-              expect(Number(firstVal)).to.be.greaterThan(Number(secondVal));
+              expect(Number.parseInt(firstVal)).to.be.greaterThan(
+                Number.parseInt(secondVal)
+              );
             });
         });
       break;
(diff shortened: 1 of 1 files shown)


pierregee (approved):

mobile-app/cypress/e2e/functional/wallet/loans/loans.spec.ts
```diff
@@ -42,7 +42,9 @@ function checkTokensSortingOrder(
             .invoke("text")
             .then((text) => {
               const secondVal = text.split("% Interest")[0];
-              expect(Number(firstVal)).to.be.lessThan(Number(secondVal));
+              expect(Number.parseInt(firstVal)).to.be.lessThan(
+                Number.parseInt(secondVal)
```
lykalabrada, line 46: Hello, I happen to be working on loans e2e as well and saw similar issue here. This change will just parse numbers like `1,234` to `1`. 
I think what we need to address is the comma inside string before parsing into a number:
`Number("1,234") -> NaN`

G9000: Hi @lykalabrada thanks for pointing out. But from my investigation, it does work to some extent. But let me know if you found much more accurate implementation. I also not so sure
Screenshot 2022-10-26 151108
Screenshot 2022-10-26 151302

lykalabrada: Hmm yes this solution works for our current use case, but if in the future the data changed and we somehow got something like values `5 ≈ 5` and `1,000 ≈ 1` then this will fail.
I'm thinking, since having comma inside the value is the root cause of this NaN issue:
image

Then maybe we can remove the comma and parse to number:
```suggestion
              const value = text.split("$")[1];
              const secondVal = value.replace(/,/g, "");
              expect(Number(firstVal)).to.be.lessThan(Number(secondVal));
```

P.S. Issue is intermittent on current main branch. I don't know why but sometimes the value got commas and sometimes it doesn't have😖

G9000: @lykalabrada tested yes that will work too. Anyway since you working on the loans e2e perhaps we could close this PR  and combine it with yours since it is a small adjustment on the files.

PR #3729, opened 2022-10-27
fix(ui-ux): remove decimalScale for NumberFormat

veralygit:
This PR is to remove all decimalScale in NumberFormat and use BigNumber(value).toFixed()
this is because the decimalScale prop cant be customized and always round the values down
using BigNumber(value).toFixed() will enable consistent rounding and decimal places

Fixes #
1. Removed decimalScale in NumberFormats. For NumberFormats with different decimalScale dp and value dp toFixed() will be using the smaller value dp 
2. fix linting issues by removing unused props/variables and adding ListRenderItemInfo<> to resolve FlatList prop type error

#### Additional comments?:

Some decimalScale removal (eg NumberRowV2) needs a larger reformat due to if decimalScale prop or value prop dp is lower. The lower dp will take precedence. Some $ value and token value are sharing the same NumberFormat so it is difficult to fix the dp to 8 or 2 as the dp differs based on $ value or token value 

For `InfoRow.tsx, InputHelperText.tsx, NumberRow.tsx, BidHistoryItem.tsx, InfoSection.tsx, NumberRowWithConversion.tsx, VaultSectionTextRowV2.tsx, NetworkDetails.tsx, NumberRowV2.tsx, VaultSectionTextRowV2.tsx` unable to verify on the UI toFixed() the dp or if they have the same behaviour as NumberRowV2. Keeping the decimalScale in these files  

#### Developer Checklist:

- [ ] Unit tests*
- [ ] Added e2e tests*
- [ ] Added translations*


mobile-app/app/screens/AppNavigator/screens/Loans/components/LoanCards.tsx
```diff
@@ -572,10 +574,9 @@ function LoanCard({
             </ThemedText>
           </View>
         )}
-        value={currentPrice}
+        value={getPrecisedCurrencyValue(currentPrice)}
```
kyleleow, line 577: let's revert this change since it has been applied in line 550

mobile-app/app/screens/AppNavigator/screens/Portfolio/screens/ConvertScreen.tsx
```diff
@@ -481,7 +478,7 @@ function ConversionResultCard(props: {
             </ThemedTextV2>
           )}
           thousandSeparator
-          value={props.totalTargetAmount}
+          value={BigNumber(props.totalTargetAmount).toFixed(8)}
```
kyleleow, line 481: let's revert this, rounding has been applied at parent component

mobile-app/app/screens/AppNavigator/screens/Loans/components/LoanCards.tsx
veralygit: reverted

lykalabrada (approved):

kyleleow (approved):

mobile-app/cypress/e2e/functional/wallet/auctions/auctions_bid_with_multiple_wallet.spec.ts
```diff
@@ -7,6 +7,8 @@ import {
   generateBlockUntilLiquidate,
 } from "../../../../support/walletCommands";
 
+BigNumber.set({ ROUNDING_MODE: BigNumber.ROUND_DOWN });
```
thedoublejay, line 10: I think there's a way for cypress to run this before all tests. Not sure

thedoublejay: Or does it makes sense to add it on all files?

thedoublejay (changes requested):

mobile-app/cypress/e2e/functional/wallet/auctions/auctions_bid_with_multiple_wallet.spec.ts
veralygit: Was unable to find a way for cypress to run the config at the start. Tested the configurations if is not added to the files BigNumber would not round down

mobile-app/App.tsx
```diff
@@ -67,6 +68,8 @@ export default function App(): JSX.Element | null {
     wallet_toast: (toast: ToastProps) => <WalletToast toast={toast} />,
   };
 
+  BigNumber.set({ ROUNDING_MODE: BigNumber.ROUND_DOWN });
```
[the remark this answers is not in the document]
veralygit: updated

PR #3781, opened 2022-11-21
feature(core): update feature flag

G9000:
This PR to update the feature flag by adding a new type to enable app distinction between Mobile LW, Desktop LW, Scan etc.

Fixes #

#### Developer Checklist:

- [ ] Unit tests*
- [ ] Added e2e tests*
- [ ] Added translations*

### DIFF
--- a/mobile-app/app/contexts/FeatureFlagContext.tsx
+++ b/mobile-app/app/contexts/FeatureFlagContext.tsx
@@ -81,6 +81,7 @@ export function FeatureFlagProvider(
     return featureFlags.some((flag: FeatureFlag) => {
       if (
         flag.networks?.includes(network) &&
+        flag.app?.includes("MOBILE_LW") &&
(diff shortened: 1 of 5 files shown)


mobile-app/app/contexts/FeatureFlagContext.tsx
```diff
@@ -81,6 +81,7 @@ export function FeatureFlagProvider(
     return featureFlags.some((flag: FeatureFlag) => {
       if (
         flag.networks?.includes(network) &&
+        flag.app?.includes("MOBILE_LW") &&
```
pierregee, line 84: Just realized this would be a breaking change for previous versions of LW, do we need to handle it? Maybe something like `flag.app === undefined || flag.app?.includes("MOBILE_LW")` CC: @thedoublejay

thedoublejay (approved):

PR #3788, opened 2022-11-22
Upgrading Tailwind CSS

### LINKED ISSUE #1968: upgrade to `tailwind@3.0.0`

platno:
It upgrades Tailwind CSS and Tailwind RN

[Fixes #1968](https://github.com/DeFiCh/wallet/issues/1968)

#### Additional comments?:
Still Work in progress, using this PR to share approach, so we can comment it.
Need to fix getting colors, fonts and some unit tests.

#### Developer Checklist:

- [ ] Unit tests*
- [ ] Added e2e tests*
- [ ] Added translations*


mobile-app/app/tailwind.ts
```diff
@@ -45,30 +45,27 @@ const fonts = {
...
+  tailwind: (...args: Argument[]) => { [key: string]: string },
+  getColor: (colorClassName: string) => string;
+} {
+  const tailwind = useTailwind();
+  
+  const getStyles = (...args: Argument[]) => {
+    return tailwind(classNames(args));
+  };
 
-/**
```
thedoublejay, line 56: Can we keep the comments here? It helps for new joiners

package.json
```diff
@@ -18,7 +18,9 @@
     "cypress:open": "cypress open",
     "cypress:run": "cypress run --headless --browser chrome",
     "react-devtools:open": "react-devtools",
-    "translation:missing": "ts-node shared/translations/reporter/index.ts"
+    "translation:missing": "ts-node shared/translations/reporter/index.ts",
+    "build:tailwind": "cd ./mobile-app/app && tailwindcss --input ./input.css --output ./tailwind.css --no-autoprefixer && tailwind-rn --input ./tailwind.css --output ./tailwind.json",
+		"dev:tailwind": "cd ./mobile-app/app && concurrently \"tailwindcss --input ./input.css --output ./tailwind.css --no-autoprefixer --watch\" \"tailwind-rn --input ./tailwind.css --output ./tailwind.json --watch\""
```
thedoublejay, line 23: Formatting problems

thedoublejay (changes requested):
Approach looks great. Nice work. 

Some comments:
1. This migration will break this CI action. I think we can just remove this. This validation runs to check if the generated styles.json is the same when generating it from CI. To prevent differences that might cause bugs. - https://github.com/DeFiCh/wallet/actions/runs/3520506533/jobs/5918488212

2. Need to update the PR title to follow semantic commits, you may refer to other PRs

3. Haven't personally tested it. Will wait for you to complete all TODOs.

mobile-app/app/tailwind.ts
```diff
@@ -1,74 +1,55 @@
...
- * Please do not edit this file.
- *
- * If you want to use extend tailwind, please edit 'tailwind.config.js' and run 'npx create-tailwind-rn'
- * in this directory after you have edited the file.
- */
-const created = create({ ...styles, ...fonts });
+  const getColor = (colorClassName: string) => {
```
platno, line 33: @thedoublejay This one is little bit hack, BUT proper solution would require a lot of changes in our components, which I would recommend to do in next iteration

mobile-app/app/input.css
```diff
@@ -0,0 +1,51 @@
+@tailwind utilities;
+
+.font-light {
```
platno, line 3: @thedoublejay This one is little bit hack, BUT proper solution would require a lot of changes in our components, which I would recommend to do in next iteration

platno:
> 1. This migration will break this CI action. I think we can just remove this. This validation runs to check if the generated styles.json is the same when generating it from CI. To prevent differences that might cause bugs. - https://github.com/DeFiCh/wallet/actions/runs/3520506533/jobs/5918488212

I will check how this is done, we need to check now is tailwind.json generated properly

> 3. Haven't personally tested it. Will wait for you to complete all TODOs.

I finalised all TODOs.

mobile-app/app/input.css
thedoublejay: Can you explain a bit what is the approach of this "hack"? And what would it take for future iterations

thedoublejay:
@platno Have you tried running it on an actual mobile device? To see if it works and no crashes?

platno:
> Have you tried running it on an actual mobile device? To see if it works and no crashes?

There was one CSS issue on iPhone, showing only white screen after splash screen.
It is fixed now, so App on iPhone looks OK.
I do not have Android device to do check there.

mobile-app/app/input.css
platno: Proper fix would be adding font family and weight in config
https://tailwindcss.com/docs/font-family
https://tailwindcss.com/docs/font-weight
And to use tailwind classes in our components.

Currently we have custom style classes that combine font family and weight into one class name.

So, this "hack" is adding fonts classes into root CSS file and being able to use them without changing components.

thedoublejay: Thanks for explaining, i'll create a placeholder ticket for now - https://github.com/DeFiCh/wallet/issues/3810

PR #3889, opened 2023-01-18
fix(core): conversion prompt again on flow reset 

JoshuaC817:
To fix conversion prompt being triggered again on any transaction flow resets
Solution: by adding in estimated transaction fees into conversion prompt

Fixes #

#### Developer Checklist:

- [x] Tested on Light mode and Dark mode\*
- [ ] Unit tests\*
- [ ] Added e2e tests\*
- [ ] Added translations\*


mobile-app/app/hooks/wallet/Conversion.ts
```diff
@@ -48,6 +54,15 @@ export function useConversion(props: useConversionProps): ConversionResult {
   const reservedDFI = 0.1;
 
   useEffect(() => {
+    function setAmount(transactionFee: BigNumber) {
+      setConversionAmount(
+        amount
+          .minus(type === "utxo" ? DFIUtxo.amount : DFIToken.amount)
+          .plus(type === "utxo" ? reservedDFI : 0)
+          .plus(type === "utxo" ? transactionFee : 0)
```
kyleleow, line 62: converting to dfi token should have tx fee too right?

JoshuaC817: ah yes

JoshuaC817: but wait, if it's dfi conversion, we don't need to convert the transaction fee to utxo because transaction fee is deducting utxo not dfi

PR #3940, opened 2023-03-22
chore(ui-ux): remove BURN2 from pools

veralygit:
As titled

Fixes #

#### Developer Checklist:

- [x] Tested on Light mode and Dark mode\*
- [x] Your UI implementation visually matched the rendered design\*
- [ ] Unit tests\*
- [ ] Added e2e tests\*
- [ ] Added translations\*


kyleleow:
I am thinking about this again, should we consider filtering out the poolpair that has `status: false` instead?
I checked that currently `status: false` poolpair are the one after stock split, e.g.: `dTSLA-dDUSD/v1`, `dGOOGL-dDUSD/v1` 

And on product side we are manually filtering again `..v1` poolpair, e.g.: 
https://github.com/WavesHQ/walletkit/blob/8a3cc3d66697a0ca483c4cd29100f088ab6b104b/packages/walletkit-ui/src/store/wallet.ts#L224-L226

For backward compatibility, maybe we should by default return all pair, and accept a new param that will filter out `status: false`? Or the other way round, i.e.: default to return filtered pair, accept a new param to return unfiltered pairs

PR #3955, opened 2023-06-08
chore(e2e): skip flaky specs

Rampraesath:

- Skip failing specs both locally and in CI. Will be fixing failures 1 by 1 as discussed

Fixes #

#### Developer Checklist:

- [ ] Tested on Light mode and Dark mode\*
- [ ] Unit tests\*
- [ ] Added e2e tests\*
- [ ] Added translations\*


cypress.config.js
```diff
@@ -0,0 +1,53 @@
+const { defineConfig } = require("cypress");
+
```
thedoublejay, line 2: @Rampraesath What is this for?

Rampraesath: as per discussion i have removed this file

PR #4082, opened 2023-10-19
fix(core): handle auto-conversion of utxo to dfi token

lykalabrada:
Handle auto-conversion of UTXO into DFI token when user is sending to an EVM address, and has enough UTXO but not enough DFI token.

Fixes DFC-363

#### Developer Checklist:

- [ ] Tested on Light mode and Dark mode\*
- [ ] Unit tests\*
- [ ] Added e2e tests\*
- [ ] Added translations\*

### DIFF
--- a/mobile-app/app/screens/AppNavigator/screens/Portfolio/screens/SendScreen.tsx
+++ b/mobile-app/app/screens/AppNavigator/screens/Portfolio/screens/SendScreen.tsx
@@ -124,11 +124,11 @@ export function SendScreen({ route, navigation }: Props): JSX.Element {
   const getInputTokenType = () => {
     if (isEvmAddress) {
       return "token";
-    } else if (token?.id === "0_unified") {
-      return "utxo";
-    } else {
-      return "others";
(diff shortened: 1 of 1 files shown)


mobile-app/app/screens/AppNavigator/screens/Portfolio/screens/SendScreen.tsx
```diff
@@ -115,14 +115,27 @@ export function SendScreen({ route, navigation }: Props): JSX.Element {
...
 
+  const getInputTokenType = () => {
+    if (isEvmAddress) {
+      return "token";
+    } else if (token?.id === "0_unified") {
+      return "utxo";
+    } else {
+      return "others";
+    }
+  };
```
fullstackninja864, line 132:
```suggestion
  const getInputTokenType = () => {
    if (isEvmAddress) {
      return "token";
    }
    if (token?.id === "0_unified") {
      return "utxo";
    }
    return "others";
  };
```

